PR #7, opened 2019-01-25
Add more checks

jeanconn:
Add more checks

### DIFF
--- a/aca_preview/preview.py
+++ b/aca_preview/preview.py
@@ -412,8 +412,8 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}"""
                 'critical', f'ER bright stars: only {n_bright} stars brighter than {bright_lim}')
 
     def check_enough_guide_for_ers(self, n_required=8):
-        """
-        Warn on ERs with fewer than n_required (8) guide stars.
+        """Warn on ERs with fewer than n_required (8) guide stars.
+
         """
         obsid = float(self.obsid)
         is_OR = abs(obsid) < 38000
@@ -421,8 +421,8 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}"""
             self.add_message('critical', f'ER guides stars: only {len(self.guides)} stars')
 
     def check_guide_pos_errs(self):
-        """
-        Warn on stars with larger POS_ERR (warning at 1" critical at 2")
+        """Warn on stars with larger POS_ERR (warning at 1" critical at 2")
+
         """
         for star in self.guides:
             agasc_id = star['id']
@@ -437,10 +437,9 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}"""
                     break
 
     def check_imposters(self):
-        """
-        Warn on stars with larger imposter centroid offsets
-        """
+        """Warn on stars with larger imposter centroid offsets
 
+        """
         # Borrow the imposter offset method from starcheck
         def imposter_offset(cand_mag, imposter_mag):
                 """
@@ -465,11 +464,13 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}"""
                     break
 
     def check_too_bright_guide(self):
-        """
-        Warn on guide stars that may be too bright.
-        Set to warn with critical warn if MAG_ACA_ERR used in selection is less than 0.1 or if
-        within 3 * mag_err of the hard 5.8 limit. Set to warn with orange warn if just brighter
-        than 6.1 (should be double-checked in context of other candidates).
+        """Warn on guide stars that may be too bright.
+
+        Set to warn with critical warn if MAG_ACA_ERR used in selection is less
+        than 0.1 or if within 3 * mag_err of the hard 5.8 limit. Set to warn
+        with orange warn if just brighter than 6.1 (should be double-checked in
+        context of other candidates).
+
         """
         for star in self.guides:
             agasc_id = star['id']
@@ -477,7 +478,7 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}"""
                 self.add_message(
                     'critical',
                     f'Guide star {agasc_id} within 3*mag_err of 5.8')
-            elif (star['mag'] < 6.1) and (star['MAG_ACA_ERR'] * .01 < 0.1):
+            elif (star['mag'] < 6.1) and (star['MAG_ACA_ERR'] * 0.01 < 0.1):
                 self.add_message(
                     'critical',
                     f'Guide star {agasc_id} < 6.1 with small MAG_ACA_ERR.  Double check selection.')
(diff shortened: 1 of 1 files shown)


aca_preview/preview.py
```diff
@@ -336,3 +340,66 @@ def check_bright_guide_for_ers(self, n_bright_req=3, bright_lim=9.0):
         n_bright = np.count_nonzero(self.guides['mag'] < bright_lim)
         if not is_OR and n_bright < n_bright_req:
             self.add_message('critical', f'ER bright stars: only {n_bright} stars brighter than {bright_lim}')
+
+    def check_enough_guide_for_ers(self, n_required=8):
+        """
+        Warn on ERs with fewer than n_required (8) guide stars.
+        """
+        obsid = float(self.obsid)
+        is_OR = abs(obsid) < 38000
```
taldcroft: Make an `is_OR` property to stop this repetition.  And while you are at it might as well put in an `is_ER` property that returns `not is_OR`.  This just makes downstream checks more obvious.

aca_preview/preview.py
```diff
@@ -336,3 +340,66 @@ def check_bright_guide_for_ers(self, n_bright_req=3, bright_lim=9.0):
         n_bright = np.count_nonzero(self.guides['mag'] < bright_lim)
         if not is_OR and n_bright < n_bright_req:
             self.add_message('critical', f'ER bright stars: only {n_bright} stars brighter than {bright_lim}')
+
+    def check_enough_guide_for_ers(self, n_required=8):
+        """
+        Warn on ERs with fewer than n_required (8) guide stars.
+        """
+        obsid = float(self.obsid)
```
taldcroft: Probably makes sense to just do this float cast in `__init__`.  I can't think of a reason we need the str version at this point.

aca_preview/preview.py
```diff
@@ -336,3 +340,66 @@ def check_bright_guide_for_ers(self, n_bright_req=3, bright_lim=9.0):
...
+            imp_offset = imposter_offset(star['mag'], star['imp_mag'])
+            for limit, category in ((4.0, 'critical'),
+                                    (2.5, 'warning')):
+                if imp_offset > limit:
+                    self.add_message(category, f'Guide star {agasc_id} imposter offset {imp_offset:.1f}, limit {limit} arcsec')
+                    break
+
+    def check_too_bright_guide(self):
+        """
+        Warn on guide stars that may be too bright.
```
taldcroft: For this and other method docstrings, put a line break between the one-line method description and the details.    In general, in emacs and hit `ESC-q` to format correctly.  In this case it would be better to format this as a list like:
```
- Critical warn if MAG_ACA_ERR used in selection is less than 0.1
- Critical warn if within 3 * mag_err ...
- ..
```

aca_preview/preview.py
```diff
@@ -336,3 +340,66 @@ def check_bright_guide_for_ers(self, n_bright_req=3, bright_lim=9.0):
...
+        Set to warn with critical warn if MAG_ACA_ERR used in selection is less than 0.1 or if within 3 * mag_err of the hard 5.8 limit.
+        Set to warn with orange warn if just brighter than 6.1 (should be double-checked in context of other candidates).
+        """
+        for star in self.guides:
+            agasc_id = star['id']
+            if star['mag'] - (3 * star['mag_err']) < 5.8:
+                self.add_message('critical', f'Guide star {agasc_id} within 3*mag_err of 5.8')
+            elif (star['mag'] < 6.1) and (star['MAG_ACA_ERR'] * .01 < 0.1):
```
taldcroft: Should be `0.01`.

aca_preview/preview.py
```diff
@@ -336,3 +340,66 @@ def check_bright_guide_for_ers(self, n_bright_req=3, bright_lim=9.0):
...
+                self.add_message('critical', f'Guide star {agasc_id} < 6.1 with small MAG_ACA_ERR.  Double check selection.')
+            elif star['mag'] < 6.1:
+                self.add_message('warning', f'Guide star {agasc_id} < 6.1. Double check selection.')
```
taldcroft: I suspect this is going to be adding noise, but for now is fine while we get experience and fine-tune proseco.

jeanconn:
Thanks for the early review Tom!  I was trying to decide what would be helpful in review run by FOT (with Tuesday deadline) and what would be helpful for my review (not as deadlined).  I think these might all be useful on the FOT check.

For me, I was thinking that checking:

* for guide stars in your list of GAIA stars with large PM 
* for guide or acq stars with magnitude histories that are inconsistent with catalog mag (would require acq and guide stats)

would both also be useful but could go in a later version or a SOT-only profile or some such.  

Also, what were you thinking about tests and how much of that should get done for the deadline?

taldcroft:
The plan above sounds good.  

For immediate testing I would suggest the strategy of hacking hard-coded params to generate and confirm warnings in existing loads.  I think some infrastructure changes may be needed to support unit testing, so best to merge this once that hack-verification is done.

jeanconn:
So far I'd mostly been testing by just constructing individual catalogs that failed the checks, putting them out as single-catalog pkls, and reviewing the outputs.

taldcroft:
> So far I'd mostly been testing by just constructing individual catalogs that failed the checks, putting them out as single-catalog pkls, and reviewing the outputs.

Yes, that works even if a bit painful.

Eventually unit tests should look like:
```
# Make some stars that are tighly clustered
aca = get_aca_catalog(att=att, stars=stars, ..)
aca = ACAReviewTable(aca)  # this is the infrastructure change
aca.check_position_on_ccd()
assert aca.messages == [{'category': 'critical', 
                         'message': Guide stars all clustered within 500" radius'}]
```

aca_preview/preview.py
jeanconn: Where is the most sensible place to set the value of the new properties?

jeanconn: I guess as long as obsid is set reasonably, I don't need a place.

taldcroft: Right, as long as obsid is set then the property is just read-only whenever needed.  No need to even store it, just compute on the fly each time.

taldcroft:
@jeanconn - you should review my updates and (hopefully) merge.

jeanconn:
Ok.  I'll need to regroup a bit because I thought I had this branch and PR
to myself and had rebased and reorganized.

aca_preview/preview.py
```diff
@@ -180,6 +187,17 @@ def add_review_methods(cls, aca):
         aca.context = {}  # Jinja2 context for output HTML review
         aca.messages = []  # Warning messages
 
+    @property
+    def is_OR(self):
+        """Return ``True`` if obsid corresponds to an OR."""
+        obsid = float(self.obsid)
+        return abs(obsid) < 38000
```
jeanconn: I spoke with Steve and John about this and I think we might be better off with just obsid < 38000 .  The negative target lists are large (on order of 7000 for this cycle).

jeanconn: I'm going to merge this and open a new PR for this issue.

PR #6, opened 2019-01-25
Make aca_preview work with 4.3.x

jeanconn:
Make aca_preview work with proseco 4.3.x .

PR #8, opened 2019-01-25
Introduce compatibility code for 4.3.x

### LINKED ISSUE #5: Make aca_preview 4.3.x compatible
It looks like I've got a conflict when trying to run from a pickle against proseco 4.3.2 (checked out in my ~/git/proseco below) which is what I believe we'd be running against if we try to install this in ska3-matlab.  I don't think 4.3.x has the updates to re-instantiate the bad_stars mask?  So then I'm not sure if we want to fix in that in proseco 4.3.x or call separate custom plotting in aca_preview.

```
/proj/sot/ska/jeanproj/git/aca_preview/aca_preview/preview.py in make_starcat_plot(self)
    145         ax = fig.add_subplot(1, 1, 1)
    146         self.acqs.stars
--> 147         self.plot(ax=ax)
    148         plt.tight_layout()
    149         fig.savefig(str(outfile))

~/git/proseco/proseco/core.py in plot(self, ax)
    493         if self.acqs:
    494             stars_kwargs['stars'] = self.acqs.stars
--> 495             stars_kwargs['bad_stars'] = self.acqs.bad_stars
    496 
    497         plot_stars(attitude=self.att, catalog=self, ax=ax, **stars_kwargs)

AttributeError: 'AcqTable' object has no attribute 'bad_stars'
```

taldcroft:
This takes a different approach from #6 and adds explicit compatibility code to make `aca_preview` work with both 4.3.x and master.

Tested with both and see no diffs in HTML output and star plots are generally similar, though the different versions do have different definitions of bad stars, so diffs are expected.  In particular 4.3.x marks all stars outside the allowed region for acq as bad, while master does not apply this spatial filtering so a reviewer can spot potential acq/guide stars to use for a roll change.

```
(ska3) neptune$ diff JAN2119A{,-4.3.2}/index.html
54c54
< Proseco version: 4.4-r515-b6ec6dd
---
> Proseco version: 4.3.2-r450-a1bf4fc
```
Closes #5 .

### DIFF
--- a/aca_preview/preview.py
+++ b/aca_preview/preview.py
@@ -16,11 +16,18 @@ import proseco
 from proseco.catalog import ACATable
 from proseco.core import StarsTable
 import proseco.characteristics as CHAR
+import proseco.characteristics_guide as GUIDE
 
 CACHE = {}
 VERSION = proseco.test(get_version=True)
 FILEDIR = Path(__file__).parent
 CATEGORIES = ('critical', 'warning', 'caution', 'info')
+
+
+# Fix characterstics compatibility issues between 4.3.x and 4.4+
+if not hasattr(CHAR, 'CCD'):
+    for attr in ('CCD', 'PIX_2_ARC', 'ARC_2_PIX'):
+        setattr(CHAR, attr, getattr(GUIDE, attr))
 
 
 def preview_load(load_name, outdir=None):
--- a/aca_preview/preview.py
+++ b/aca_preview/preview.py
@@ -64,6 +64,7 @@ def preview_load(load_name, outdir=None):
         aca.outdir = outdir
         aca.context = {}
         aca.messages = []
+        aca.set_stars_and_mask()
         aca.preview()
         acas.append(aca)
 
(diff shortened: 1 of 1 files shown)


aca_preview/preview.py
```diff
@@ -126,6 +134,24 @@ def get_summary_text(acas):
...
+    def set_stars_and_mask(self):
+        """Set stars attribute for plotting.
+
+        This includes compatibility code to deal with somewhat-broken behavior
+        in 4.3.x where the base plot method is hard-coded to look at
+        ``acqs.stars`` and ``acqs.bad_stars``.
+
+        """
+        # Get stars from AGASC and set ``stars`` attribute
+        self.set_stars()
```
jeanconn: This looks good but not sure if you want to handle the different behaviors in set_stars with regard to filter_near_fov.  Is there a use case where it would make a difference?

aca_preview/preview.py
```diff
@@ -16,13 +16,20 @@
...
 import proseco.characteristics as CHAR
+import proseco.characteristics_guide as GUIDE
 
 CACHE = {}
 VERSION = proseco.test(get_version=True)
 FILEDIR = Path(__file__).parent
 CATEGORIES = ('critical', 'warning', 'caution', 'info')
 
 
+# Fix characterstics compatibility issues between 4.3.x and 4.4+
```
jeanconn: Comment typo

aca_preview/preview.py
taldcroft: I can't think of any reason to care in this context.  If we come across a driver then we'll address it then.

PR #11, opened 2019-01-25
Update infrastructure to support unit testing

### LINKED ISSUE #9: Need one-liner or entry-point script
Have not worked out details with FOT for running the tool, but a command line utility or a one-liner for a pyexec call is most likely the way to go.  I don't know if we need to get the browser call to run.

### LINKED ISSUE #10: Needs a version

taldcroft:
This makes a new class method to add the review methods to an existing ACATable object and set it up for doing checks.

It also adds an example of how to write tests of the preview checks.

Fixes #10 
Closes #9


aca_preview/tests/test_checks.py
```diff
@@ -0,0 +1,21 @@
...
+    stars = StarsTable.empty()
+    stars.add_fake_constellation(n_stars=3, mag=10.25)
+    aca = get_aca_catalog(**STD_INFO, stars=stars, dark=DARK40)
+    ACAReviewTable.add_review_methods(aca)
+
+    aca.check_acq_p2()
+    assert len(aca.messages) == 1
+    msg = aca.messages[0]
+    assert msg['category'] == 'critical'
+    assert 'less than 2.0 for OR' in msg['text']
```
taldcroft: Note that this does not include the actual value of P2 that is reported in the full message, just that some bit of the message matches expectation.  This future-proofs against slight changes due to model updates etc.

taldcroft:
Since the entry point script counts as infrastructure I glopped that into this PR.

aca_preview/preview.py
```diff
@@ -30,7 +35,26 @@
...
 
-def preview_load(load_name, outdir=None):
+def main(sys_args=None):
+    """Command line interface to preview_load()"""
+
+    import argparse
+    parser = argparse.ArgumentParser(description='ACA preliminary review tool')
+    parser.add_argument('load_name',
+                        type=str,
+                        help='Load name (e.g. JAN2119A)')
```
jeanconn: This looks good.  Me, I prefer filename for better natural tab completion and being able to call on a pickle conveniently outside my working directory, but 'load_name' may continue to make the most sense for product directories.

aca_preview/tests/test_checks.py
```diff
@@ -0,0 +1,21 @@
...
+from proseco.tests.test_common import DARK40, STD_INFO
+from aca_preview.preview import ACAReviewTable
+
+
+def test_check_P2():
+    """Test the check of acq P2"""
+    stars = StarsTable.empty()
+    stars.add_fake_constellation(n_stars=3, mag=10.25)
+    aca = get_aca_catalog(**STD_INFO, stars=stars, dark=DARK40)
+    ACAReviewTable.add_review_methods(aca)
```
jeanconn: OK, so since the aca is somewhat converted to the ACAReviewTable, it is never pickled so the stars stay in the structure and set_stars doesn't fetch from the agasc?  Just trying to follow this through.

aca_preview/preview.py
taldcroft: Fair enough.  At risk of over-complicating things, I added 1bb7278 for your tab-completion pleasure.

aca_preview/tests/test_checks.py
taldcroft: Here `aca` is fully converted to `ACAReviewTable` (albeit by a tricky pathway).  There is no pickling and the `set_stars()` method is never called.  That only happens in the main program because in that case the original `ACATable` is from a pickle and doesn't already have stars.

taldcroft:
@jeanconn - feel free to merge if you are happy with this.

PR #13, opened 2019-01-27
Make guide and acq detail reports and include links

taldcroft:
This is a start.

To do:
- [x] Put links in to the index.html report

### DIFF
--- a/aca_preview/preview.py
+++ b/aca_preview/preview.py
@@ -214,7 +214,7 @@ class ACAReviewTable(ACATable):
         # in these cases.
         if obsid is not None:
             f_obsid = round(float(obsid), 1)
-            i_obsid = int(obsid)
+            i_obsid = int(f_obsid)
             num_obsid = i_obsid if (i_obsid == f_obsid) else f_obsid
 
             aca.obsid = num_obsid
--- a/aca_preview/preview.py
+++ b/aca_preview/preview.py
@@ -604,3 +604,8 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}"""
                 self.add_message(
                     'warning',
                     f'Guide star {agasc_id} < 6.1. Double check selection.')
+
+
+# Run from source ``python -m aca_preview.preview <load_name> [options]``
+if __name__ == '__main__':
+    main()
--- a/aca_preview/preview.py
+++ b/aca_preview/preview.py
@@ -49,7 +49,7 @@ def main(sys_args=None):
                         help='Output directory (default=<load name>')
     parser.add_argument('--report-level',
                         type=str,
-                        default='none',
+                        default='warning',
                         help="Make reports for messages at/above level "
                              "('all'|'none'|'info'|'caution'|'warning'|'critical') "
                              "(default='none')")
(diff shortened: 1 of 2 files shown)


jeanconn:
Do you want to just cut f13d90c from this PR and I'll finish my reorganization of those checks and the tests for them?

jeanconn:
Oh nevermind, I'm guessing this is based on the other PR.  Fine I'll just do everything as new commits.

taldcroft:
Sorry, should have mentioned this being based on #7.  This will be a trivial rebase once that gets merged.  There was a reason for basing on #7 though I don't recall it now.

taldcroft:
I'm going to propose that the default `--report-level` is `warning`.  That should cover most cases where inspection of the details is helpful.  Thoughts @jeanconn?

taldcroft:
@jeanconn - I've tested this with a local version of 4.3.3 with all the 4.3.3 milestoned proseco PR's merged and the JAN2819P prelim pickle.  It seems to basically work, but I haven't checked on details.  Please have a look with your usual attention to detail.  :smile:

aca_preview/preview.py
```diff
@@ -167,18 +192,90 @@ def get_summary_text(acas):
...
         aca.messages = []  # Warning messages
+        aca.loud = loud
+
+        # Input obsid could be a string repr of a number that might have have
+        # up to 1 decimal point.  This is the case when obsid is taken from the
+        # ORviewer dict of ACATable pickles from prelim review.  Tidy things up
+        # in these cases.
+        if obsid is not None:
+            f_obsid = round(float(obsid), 1)
+            i_obsid = int(obsid)
```
jeanconn:
```
ine 217, in add_review_methods
    i_obsid = int(obsid)
ValueError: invalid literal for int() with base 10: '20208.1'
```
Also, what's the best way to test entry-point scripts?  I'm used to just having a __name__ == main in the source script to run it from the source dir.

aca_preview/preview.py
```diff
@@ -167,18 +192,90 @@ def get_summary_text(acas):
...
+        :param loud: print processing status info (default=False)
 
         """
         aca.__class__ = cls
         aca.context = {}  # Jinja2 context for output HTML review
         aca.messages = []  # Warning messages
+        aca.loud = loud
+
+        # Input obsid could be a string repr of a number that might have have
+        # up to 1 decimal point.  This is the case when obsid is taken from the
```
jeanconn: Also, I think this "up to 1 decimal point" may be wrong.  For the examples we've seen, sure that .19999 was a poor representation of .2, but it does not seem impossible to me that we'd have 10 or more splits of a single target in a prelim (2 or 3 seems much more likely I agree but seems not worth breaking over more).

aca_preview/preview.py
jeanconn: I'm guessing you just intended
```
i_obsid = int(f_obsid)
```

aca_preview/preview.py
taldcroft: I set this to 2 decimal points.  I can't immediately find your example of the MATLAB obsid with a poor representation though.

jeanconn: You are welcome to the early FEB0419 pickle in ASPECT_ICXC/SFE/2019/FEB0419/FEB0419_Test_proseco.pkl .

aca_preview/preview.py
```diff
@@ -47,15 +47,22 @@ def main(sys_args=None):
     parser.add_argument('--outdir',
                         type=str,
                         help='Output directory (default=<load name>')
+    parser.add_argument('--report-level',
+                        type=str,
+                        default='warning',
+                        help="Make reports for messages at/above level "
+                             "('all'|'none'|'info'|'caution'|'warning'|'critical') "
+                             "(default='none')")
```
jeanconn: Help text out of date with updated default

aca_preview/preview.py
taldcroft: Passes formatting test, and provided inspiration for the next PR.

PR #14, opened 2019-01-28
Update the guide star checks and add tests

jeanconn:
Update the guide star checks and add tests


aca_preview/preview.py
```diff
@@ -184,6 +184,7 @@ def add_review_methods(cls, aca):
 
         """
         aca.__class__ = cls
+        aca.add_row_col()
```
jeanconn: This may be controversial.  I mostly just wanted it for the tests, but then I didn't see any downside to guaranteeing that stars had row and column if they got into an ACAReviewTable.

aca_preview/preview.py
```diff
@@ -457,52 +458,59 @@ def check_enough_guide_for_ers(self, n_required=8):
...
-            for limit, category in ((2.0, 'critical'),
-                                    (1.0, 'warning')):
-                if pos_err > limit:
-                    self.add_message(
-                        category,
-                        f'Guide star {agasc_id} has POS_ERR {pos_err:.2f}, limit {limit} arcsec')
-                    break
+        if entry['id'] not in self.guides['id']:
+            return
+        star = self.guides[self.guides['id'] == entry['id']][0]
```
jeanconn: This is awkward and repeated code, but I didn't see a something natural that made more sense.  I honestly couldn't remember why the guide table seems to have an 'idx' but that 'idx' does not match the catalog table version, so ended up needing to re-match-things-up for each guide star as checked.

jeanconn:
These tests could all be better and more dynamic, but at least these are providing some coverage.

aca_preview/__init__.py
```diff
@@ -1 +1,8 @@
 __version__ = '4.0'
+
+def test(*args, **kwargs):
+    '''
```
taldcroft: Use `"""` always.

aca_preview/preview.py
taldcroft: Looks good to me

aca_preview/preview.py
taldcroft: There is a "natural" right way to do this, with the `get_id()` method: `star = self.guides.get_id(entry['id'])`.  This is really powerful so remember to use this in proseco-related code.  It works for any of the tables including `StarsTable`.

For the `aca.acqs` table at least, the `idx` column is reserved for the index into the `cand_acqs` table.  I don't remember how guide works.

aca_preview/tests/test_checks.py
```diff
@@ -14,8 +14,104 @@ def test_check_P2():
...
+    msg = aca.messages[0]
+    assert msg['category'] == 'critical'
+    assert 'Guide star imposter offset' in msg['text']
+
+
+def test_too_bright_guide_magerr():
+    """Test the check for too-bright guide stars"""
+    stars = StarsTable.empty()
+    # Add two stars because separate P2 tests seem to break with just one star
+    stars.add_fake_star(id=100, yang=100, zang=-200, mag=6.0, mag_err=.1)
```
taldcroft: 0.1

jeanconn:
What happened to give me a few minutes? :smile:

taldcroft:
> What happened to give me a few minutes?

Oops.  Didn't see the "hold off" message.

aca_preview/__init__.py
jeanconn: Oops, but I copied this from proseco which at some point want a fix then
https://github.com/sot/proseco/commit/f7a5a383757f4c3eee50c8fb9238d5964d1b57e3

aca_preview/preview.py
jeanconn: Yes.  Thanks!  I'd forgotten about this but knew there was something better.  And then saw in your #18 and fixed.  Though I'm still not sure if it makes sense to loop over the all entries of the star catalog and only pass guide stars to these guide methods or if it makes sense to pass the entries to the methods and let them reject non-guide items.  Either way, I thought having the individual guide tests run on single items made more sense from a somewhat atomic / testing perspective.

jeanconn:
Thanks @taldcroft  I think I'm done with reorganization and force-pushing with abandon in response to #18 if you'd like to take another look and see if this is going where you'd like it to go.

jeanconn:
Oops again, looks like my tests were OK with aca_preview.test() but are failing with pytest in the directory so I'll need to fix that too.

jeanconn:
Or all my tests could be broken at this juncture and just looked like they were passing... Gah..

jeanconn:
OK.  This looks like an issue with the lack of a relative import in test_checks.py and lack of an `__init__py` in aca_preview.tests, which I've added.

aca_preview/preview.py
```diff
@@ -457,71 +460,81 @@ def check_enough_guide_for_ers(self, n_required=8):
...
+        if (star['mag'] < 6.1) and (mag_aca_err < 0.1):
+            self.add_message(
+                'critical',
+                f'Guide star {agasc_id} < 6.1 with small MAG_ACA_ERR={mag_aca_err:.2f}.  '
+                f'Double check selection.',
+                idx=idx)
+        elif star['mag'] < 6.1:
+            self.add_message(
+                'warning',
+                f'Guide star {agasc_id} < 6.1. Double check selection.', idx=idx)
```
jeanconn: It is likely that a star that hits the mult*mag_err will also hit this one, but I'm OK with one critical warning and one plain warning in that case.

jeanconn:
Rebased after merging #13

aca_preview/preview.py
taldcroft: > Though I'm still not sure if it makes sense to loop over the all entries of the star catalog and only pass guide stars to these guide methods

Yes, I would write a top-level `if` statement in the loop and only pass guide entries to the routines.  That reduces repetition in the check methods themselves.  I actually started writing a comment to this effect but then went with the perfect is the enemy of good mantra.  So your call at this point.

taldcroft: And yes, I agree very much with making the tests atomic and work on just a single entry if that is the test.  I was quite happy to see that change.

aca_preview/__init__.py
taldcroft: Gahh.  I don't know how I could have done that since it really bothers me when I see `'''`.

aca_preview/preview.py
```diff
@@ -391,14 +392,16 @@ def check_catalog(self):
         """
         for entry in self:
             self.check_guide_fid_position_on_ccd(entry)
+            if entry['id'] in self.guides:
```
jeanconn: Right so I think I have "Yes, I would write a top-level if statement in the loop and only pass guide entries" done with this change, but this should probably be `if entry['id'] in self.guides['id']`

aca_preview/__init__.py
jeanconn: Looked like it was in pyyaks when I was just over there too, so I have a feeling they have a common ancestor (and that could be my fault)...

aca_preview/preview.py
taldcroft: I had the same thought and the same "I'm OK" for now.  I hope to make these tests tighter with experience (and more time to think about it).

aca_preview/preview.py
taldcroft: Yup.

jeanconn: Fixed.  Though surprised the previous code didn't error or warn.

aca_preview/tests/test_checks.py
```diff
@@ -4,7 +4,7 @@
 from proseco import get_aca_catalog
 from proseco.core import StarsTable
 from proseco.tests.test_common import DARK40, STD_INFO, mod_std_info
-from aca_preview.preview import ACAReviewTable
+from ..preview import ACAReviewTable
```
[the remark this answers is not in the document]
jeanconn: It looks like this was a case of me making two changes at once and deciding they were both fine instead of confirming they were both required.  This was how we had testing set in proseco, so I went with it when I saw something broken.  The something that was broken was really the lack of the `__init__py` file in the tests directory.  Without that `__init__.py` , py.test was, I think, loading both the wrong version of the tests and the wrong version of the code installed in my local conda env.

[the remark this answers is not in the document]
jeanconn: Since it works relative or absolute and  @taldcroft mentions being on the fence, I'll plan to merge this as-is, but we can always swap it back in a tiny change.

PR #15, opened 2019-01-28
Define an OR as an obsid less than 38000

jeanconn:
Define an OR as an obsid less than 38000


jeanconn:
I don't feel passionate about this PR, but have no reason to believe at this time that an observation with obsid -50000 would be an ER.  Probably 4 years before we'd care.

aca_preview/preview.py
```diff
@@ -245,7 +245,8 @@ def add_review_methods(cls, aca, *, obsid=None, loud=False, preview_dir='.'):
     @property
     def is_OR(self):
         """Return ``True`` if obsid corresponds to an OR."""
-        return abs(self.obsid) < 38000
+        obsid = float(self.obsid)
```
jeanconn: I don't recall how master ended up *without* the float cast, but I think it is still OK for this PR?

jeanconn: Ah, OK, so I think 13d3209 guarantees that self.obsid will be either an int or a float, so I don't need the cast in the property.

PR #18, opened 2019-01-28
Check tweaks

taldcroft:
This makes a few tweaks to checks, see 20e6a01.  The main thing is downgrading the ``3*mag_err`` check to warning, and adding critical for ``2*mag_err``.

This is based on #13 (so I could easily run and test and get nice outputs).


jeanconn:
I think 20e6a01 will conflict with #14 , so perhaps we could cut that from here and put that in there?

taldcroft:
> I think 20e6a01 will conflict with #14 , so perhaps we could cut that from here and put that in there?

Sounds like a plan, so why don't you do that by whatever means works (probably just by-hand figuring out the right edits).

jeanconn:
OK.  Hold off on review of #14 and I'll reorganize it a bit again.

PR #31, opened 2019-01-30
Add pitch_rolldev.csv to package data

jeanconn:
Add pitch_rolldev.csv to package data


jeanconn:
I've forgotten what include_package_data is supposed to help with but this seems to at least include the csv file in the package in local testing.

jeanconn:
Sorry I didn't catch it earlier.  I started off with testing from an installed aca_preview but for the roll stuff was so rushed I was working from the repo.

PR #35, opened 2019-02-04
Check for stars with missing PM in AGASC large PM GAIA

jeanconn:
Check for stars with missing PM in AGASC large PM GAIA

PR #39, opened 2019-02-07
Experiment with using an additional agasc file for bad stars

jeanconn:
Experiment with using  a new h5 file that includes the columns from proseco agasc file plus BAD_GUIDE and BAD_ACQ.  

Reimplements the GAIA PM PR #35  just warning on guide stars found in the new file with BAD_GUIDE != 0.

We'd lose insight into the "why" if we don't map BAD_GUIDE integers to codes or some such, but I think that is unlikely to help us.

### DIFF
--- a/aca_preview/preview.py
+++ b/aca_preview/preview.py
@@ -4,6 +4,7 @@
 """
 Preliminary review of ACA catalogs selected by proseco.
 """
+import os
 import io
 import re
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


taldcroft:
Conceptually nice, but...

Unfortunately we are now in a position where performance matters.  I think this approach is significantly slower than it could be for what it's doing.  Every star in the schedule causes an open / read_where / close of the supplemental catalog.

jeanconn:
Speaking of speed, I assume `agasc` als needs a fix in general to stop opening and making the default cache just on import (if I'm reading https://github.com/sot/agasc/blob/master/agasc/agasc.py#L67 correctly) especially since we're not using the miniagasc in proseco.

taldcroft:
> Speaking of speed, I assume `agasc` als needs a fix in general to stop opening and making the default cache just on import

That code only sets the file name on import but does not open anything until an `ra` or `dec` attribute is requested.

PR #38, opened 2019-02-06
Update guide_count API; match current threshold for 3 x 9.0 mag stars

taldcroft:
Addresses https://github.com/sot/aca_preview/issues/37#issuecomment-461176243.

Issue #37, opened 2019-02-06
Use fractional count of 9th mag stars > 3.0 criterion for ERs


taldcroft:
The FDB checklist requirement is:
ACA-046 | Magnitude | X | X | Perigee catalogs have 3 or more GS magnitude 9.0 or brighter

That was written forever-ago and does not consider temperature dependence.  I think we can interpret magnitude to be "temperature-corrected" magnitude and "3 or more" to be the fractional count.

jeanconn:
Agreed.  Though not quite clear on interpret ... "3 or more" to be the fractional count?  I assume that just means to set a check threshold on fractional count to about 3.  It also looks like using the ER adjusted fractional count would be significantly more conservative that the the current guidance (and I'm fine with more conservative, just keeping visibility on it).

In [53]: guide_count([9.0, 9.0, 9.0, 10.3, 10.3, 10.3], -12.96, is_ER=True)
Out[53]: 2.9978860255561708

I was trying to keep this at mag=9.0 for the quick release because planners wouldn't necessarily be able to see they had satisfied the fractional count (or how to satisfy it), but with the next planned release red and orange warnings should be visible to them and I won't have that concern.

taldcroft:
Good point.  My intent was to match current behavior so I was off by one index.  The new behavior is slightly less conservative, except that we will mostly be planning at temps above -10.9C so it probably washes out.  See #38.

jeanconn:
Done in #40

Issue #36, opened 2019-02-06
Change ER guide stars: only 6 stars from critical to warning


taldcroft:
The message "ER guide stars: only 6 stars" is critical.  Maybe this should be a warning (preference) since it does meet the ACA-008 checklist requirement.

jeanconn:
Maybe?  I coded these up quickly wondering if at this point in the mission we really did have an implicit requirement that ERs have 8 guide stars (in addition to the explicit require that there are at least 3 9th or brighter ones).  Fine to demote a 6 or 7 guide star case to warning.  Would we critical warn on 5 guide stars? (doesn't really meet checklist requirements at that point)

taldcroft:
Yes, critical at 5, warn at 6.

PR #40, opened 2019-02-07
Update guide count (n and fractional) checks

jeanconn:
This is just some quick updates for #37 and #36 .  Will need a few tweaks after #38 .

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -715,16 +715,17 @@ Predicted Acq CCD temperature (init) : {self.acqs.t_ccd:.1f}"""
         """Check for sufficient guide star fractional count
 
         """
+        obs_type = 'ER' if self.is_ER else 'OR'
         if self.is_ER and self.guide_count_9th < 3.0:
             self.add_message(
                 'critical',
-                f'ER guide count: only {self.guide_count_9th} bright stars for mag')
-        obs_type = 'OR' if self.is_OR else 'ER'
+                f'{obs_type} count of 9th mag guide stars {self.guide_count_9th:.2f} < 3.0')
+
         count_lim = 4.0 if self.is_OR else 6.0
         if self.guide_count < count_lim:
             self.add_message(
                 'critical',
-                f'Guide count: {self.guide_count} < {count_lim} for {obs_type}')
+                f'{obs_type} count of guide stars {self.guide_count:.2f} < {count_lim}')
 
     def check_pos_err_guide(self, star):
         """Warn on stars with larger POS_ERR (warning at 1" critical at 2")
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -323,7 +323,6 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         self.obsid_dir = None
         self.roll_options_table = None
         self.acq_count = None
-        self.guide_count = None
         self._is_OR = None
 
         # Input obsid could be a string repr of a number that might have have
@@ -339,6 +338,11 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
             self.acqs.obsid = num_obsid
             self.guides.obsid = num_obsid
             self.fids.obsid = num_obsid
+
+        # Compute guide count once for the record
+        self.guide_count = guide_count(self.guides['mag'], self.guides.t_ccd)
+        if self.is_ER:
+            self.guide_count_9th = guide_count(self.guides['mag'], self.guides.t_ccd, self.is_ER)
 
         if 'mag_err' not in self.colnames:
             # Add 'mag_err' column after 'mag' using 'mag_err' from guides and acqs
@@ -517,7 +521,6 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         self._base_repr_()  # Hack to set default ``format`` for cols as needed
         catalog = '\n'.join(self.pformat(max_width=-1))
         self.acq_count = np.sum(self.acqs['p_acq'])
-        self.guide_count = guide_count(self.guides['mag'], self.guides.t_ccd, self.is_ER)
 
         message_text = self.get_formatted_messages()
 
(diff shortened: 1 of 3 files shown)


aca_preview/preview.py
```diff
@@ -325,6 +325,9 @@ def add_review_methods(cls, aca, *, obsid=None, loud=False, preview_dir='.'):
             obj.t_ccd = round(obj.t_ccd, 2)
         aca.acqs.man_angle = round(obj.man_angle, 2)
 
+        # Compute guide count once for the record
+        aca.guide_count = guide_count(aca.guides['mag'], aca.guides.t_ccd, aca.is_ER)
```
taldcroft: For ER's we're going to want both `aca.guide_count_9th` and the classic `aca.guide_count`.

aca_preview/preview.py
```diff
@@ -520,7 +522,7 @@ def check_catalog(self):
 
         self.check_guide_geometry()
         self.check_acq_p2()
-        self.check_bright_guide_for_ers()
+        self.check_guide_fraction()
```
taldcroft: Try to maintain consistency with existing conventions.  In this case we always use "guide_count".  For me `guide_fraction` is a fraction like 0.77.

aca_preview/preview.py
```diff
@@ -648,21 +650,31 @@ def check_acq_p2(self):
...
 
-    def check_bright_guide_for_ers(self, n_bright_req=3, bright_lim=9.0):
-        """Check for at least 3 guide stars brighter than 9th mag for ERs.
+    def check_guide_fraction(self):
+        """Check for sufficient guide star fractional count
 
         """
-        n_bright = np.count_nonzero(self.guides['mag'] < bright_lim)
-        if self.is_ER and n_bright < n_bright_req:
+        if self.is_ER and self.guide_count < 3.0:
```
taldcroft: Should be `self.guide_count_9th < 3.0` per previous comment.

aca_preview/preview.py
```diff
@@ -648,21 +650,31 @@ def check_acq_p2(self):
...
-    def check_enough_guide_for_ers(self, n_required=8):
-        """Warn on ERs with fewer than n_required (8) guide stars.
+    def check_enough_guide_for_ers(self):
+        """Warn on ERs with fewer than required or desired number of guide stars.
 
         """
-        if self.is_ER and len(self.guides) < n_required:
-            self.add_message('critical', f'ER guide stars: only {len(self.guides)} stars')
+        if self.is_ER:
+            for n, category in ((6, 'critical'),
```
taldcroft: Actually what probably makes sense here and is consistent with ORs is to just have one critical warning for `self.guide_count < 6.0` (where this is the version called with `count_9th=False`).  Then just cut this whole check.

The only question is precedent.  This standard may be higher than what we've been planning, not sure.

aca_preview/preview.py
jeanconn: For me, sticking to convention here was actually confusing.  This is not the guide_count of starcheck or the count of the number of guide stars in the catalog.  I'd do better with a new name, but OK.

aca_preview/preview.py
jeanconn: OK.  That would simplify.

aca_preview/preview.py
jeanconn: That makes sense for the checks for me, no problem, but I was hesitant to change any of your stuff.  So I thought the bright ER guide count was already being used instead of guide_count  for the printing and the roll checks.

taldcroft: > I thought the bright ER guide count was already being used

It was, and as we saw that was really confusing.

aca_preview/preview.py
```diff
@@ -650,31 +651,20 @@ def check_acq_p2(self):
...
         """Check for sufficient guide star fractional count
 
         """
-        if self.is_ER and self.guide_count < 3.0:
+        if self.is_ER and self.guide_count_9th < 3.0:
             self.add_message(
                 'critical',
-                f'ER guide count: only {self.guide_count} bright stars for mag')
-        if self.is_OR and self.guide_count < 4.0:
+                f'ER guide count: only {self.guide_count_9th} bright stars for mag')
```
taldcroft: How about the message `{obs_type} count of 9th mag guide stars {self.guide_count_9th:.2f} < 3.0`.  Then the corresponding message below is `{obs_type} count of guide stars {self.guide_count:.2f} < {count_lim}`

aca_preview/tests/test_checks.py
```diff
@@ -34,66 +34,56 @@ def test_check_P2():
...
     aca = get_aca_catalog(**mod_std_info(n_fid=0, n_guide=8), stars=stars, dark=DARK40,
                           raise_exc=True)
     ACAReviewTable.add_review_methods(aca, obsid=50000)
-    aca.check_guide_fraction()
+    aca.check_guide_count()
     assert len(aca.messages) == 1
     msg = aca.messages[0]
     assert msg['category'] == 'critical'
-    assert 'ER guide count' in msg['text']
+    assert 'Guide count' in msg['text']
```
taldcroft: These messages should be very stable, so test the entire thing (including the output numbers).  As written it is difficult to tell what's going on.

aca_preview/tests/test_checks.py
```diff
@@ -34,66 +34,56 @@ def test_check_P2():
...
-
+    assert msg['category'] == 'critical'
+    assert 'ER guide count' in msg['text']
 
-def test_guide_fraction_er():
-    """Test the check that an OR or ER has enough fractional guide stars by guide_count"""
+    # This configuration should have not enough stars overall
     stars = StarsTable.empty()
-    stars.add_fake_constellation(n_stars=8, mag=9.5)
+    stars.add_fake_constellation(n_stars=3, mag=[8.5, 8.5, 8.5])
```
taldcroft: You should have test cases that are not slam dunks, but relatively close to the limits.  That's straightforward to do by having values right at 1.0 (9.0 mag or 10.0 mag) and bumping the temperature 0.2 C above or below -10.9.

aca_preview/preview.py
taldcroft: Fixed.

aca_preview/tests/test_checks.py
taldcroft: Fixed.

aca_preview/tests/test_checks.py
taldcroft: OK.

PR #54, opened 2019-02-16
Add test for star in proseco bad star set

taldcroft:
This is a simple version to check for stars in the bad star list.  It does not report the source of the bad star, but I figure this is going to happen very rarely.  We can come back to this later on.

It will conflict with #53, so I'll fix that up.

This includes an unrelated change to call the characterstics.py as ACA not CHAR, and some auto-code reformatting from PyCharm.

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -12,6 +12,7 @@ import pickle
 from itertools import combinations, chain
 
 import matplotlib
+
 matplotlib.use('Agg')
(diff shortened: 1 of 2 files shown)


taldcroft:
> Would there be value in having a separate acq version of the test?

If we ever see a need for a distinct acq vs. guide bad star sets then a different check here will be natural.

PR #55, opened 2019-02-17
Tweak guide/fid edge check and add tests

### LINKED ISSUE #41: Should we keep orange warning for guide stars within 5 pixels of the padded edge?
Proseco selection will fix the critical warnings for within 2.5 pixels of the edge because selection will be forbidden within 3 pixels of the padded edge.  Then does the orange warning on < 5 really have value?  Should it really drive star edit or roll change as a preference?

taldcroft:
Closes #41

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -691,8 +691,8 @@ Predicted Acq CCD temperature (init) : {self.acqs.t_ccd:.1f}"""
         if entry_type in ('GUI', 'BOT', 'FID'):
             for axis in ('row', 'col'):
                 track_delta = abs(track_lims[axis]) - abs(entry[axis])
-                for delta_lim, category in ((2.5, 'critical'),
-                                            (5.0, 'warning')):
+                for delta_lim, category in ((3.0, 'critical'),
+                                            (5.0, 'info')):
                     if track_delta < delta_lim:
                         text = (f"Less than {delta_lim} pix edge margin {axis} "
                                 f"lim {track_lims[axis]:.1f} "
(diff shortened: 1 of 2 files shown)


sparkles/core.py
```diff
@@ -688,18 +686,17 @@ def sign(axis):
...
-                    if track_delta < delta_lim:
-                        text = (f"Less than {delta_lim} pix edge margin {axis} "
-                                f"lim {track_lims[axis]:.1f} "
-                                f"val {entry[axis]:.1f} "
-                                f"delta {track_delta:.1f}")
-                        self.add_message(category, text, idx=entry['idx'])
-                        break
+        for axis in ('row', 'col'):
+            track_delta = abs(track_lims[axis]) - abs(entry[axis])
+            for delta_lim, category in ((3.0, 'critical'),
```
jeanconn: I didn't really see a need to change the critical warning to be more restrictive, but don't have a strong opinion on it.

taldcroft: I just decided to make it match proseco.  Basically I getting a star that fails this means something odd is happening (probably a manual edit, or ??).

jeanconn: I understood that.  I just figured that "odd things happening" still wasn't worthy of a critical warning in the 2.5 to 3.0 arcsec range, but this is fine.

jeanconn: I suppose I'm not sure if we can end up with a critical warning on a tiny numeric issue on sparkles vs proseco when they have exactly the same threshold.

taldcroft: That requires a selected star that is, by chance, within something like 1e-7 of the threshold.  I'm calling that in the weeds, and if it actually happens there is still no real operational impact.  I also suspect that the values will be the same at the binary level because pickling is lossless and we are generally using the same code to compute things.

jeanconn: I was thinking more about the starcheck side, but suppose we may have another sparkles release to think about this by the time we have starcheck/sparkles integration of checks.

PR #56, opened 2019-02-17
 Expand star plot and mark other roll candidates

taldcroft:
This makes plots that look like so, where stars that will be considered as roll options are circled in red.

image

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -12,9 +12,10 @@ import pickle
 from itertools import combinations, chain
 
 import matplotlib
-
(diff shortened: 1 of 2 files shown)


jeanconn:
Looks great.  Have you bumped the star filters /  bounds yet to get candidates right out to the useful allowed region for the possible roll range (the obsid 21477 in FEB1119 prelim issue)?

taldcroft:
Sparkles is now doing the right thing now for 21477 and getting that bright star at the max roll which makes it all good:

image

PR #57, opened 2019-02-18
Three fixes in the command line interface

### LINKED ISSUE #32: obsids don't seem to select into acas
For preview_load, I thought if I included an array of acas in `acas` and one obsid in `obsids` I would get the report for that obsid.  But I don't think there is any selecting into the acas in that context, so `obsids` just needs to be the list of obsids for those acas.

Probably another reason to go back to the dict eventually.

taldcroft:
- Cast --obsid as float
- Chop path prefix from input `load_name` (makes nicer title in HTML)
- As a desired side-effect that makes the auto `outdir` default to
  being in the current directory.  (@jeanconn - you were right, I was wrong.  :smile_cat:)

Fixes #32.

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -60,6 +60,7 @@ def main(sys_args=None):
                              "(default='critical')")
     parser.add_argument('--obsid',
                         action='append',
+                        type=float,
                         help="Process only this obsid (can specify multiple times, default=all")
     parser.add_argument('--quiet',
                         action='store_true',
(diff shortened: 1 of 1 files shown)


sparkles/core.py
```diff
@@ -127,7 +128,9 @@ def run_aca_review(load_name=None, *, acas=None, make_html=True, outdir=None,
         if outdir is None:
             if not load_name:
                 raise ValueError('load_name must be provided if outdir is not specified')
-            outdir = re.sub(r'(_proseco)?.pkl', '', load_name) + '_aca_preview'
+            # Chop any directory path from load_name
+            load_name = Path(load_name).name
```
jeanconn: Thanks!

PR #62, opened 2019-02-18
Make roll_options 'aca' element always be ACAReviewTable

### LINKED ISSUE #61: ACAReviewTable messages not consistent with roll_options messages
I was under the impression that the first roll_option was always going to be the original roll, however, I have found that sometimes:

```
>> acar.roll_options[0]["aca"].messages
[]
>> acar.messages
 [{'text': 'Guide star 28846800 within 3*mag_err of 5.8 (mag_err=0.10)', 'category': 'warning', 'idx': 7}, {'text': 'Guide star 28846800 < 6.1. Double check selection.', 'category': 'warning', 'idx': 7}]
```
 
Why are these different?

taldcroft:
Previously the `'aca'` key element of each roll option dict was either `ACAReviewTable` (for the first one) or `ACATable` for subsequent ones (new roll options).  Then later they were all being converted to `ACAReviewTable` and the original tossed.

This makes them all be `ACAReviewTable` from the outset and be persistent so that the messages generated in review appear on the relevant objects in the `roll_options` list.

Fixes #61

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -344,6 +344,7 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         # up to 2 decimal points.  This is the case when obsid is taken from the
         # ORviewer dict of ACATable pickles from prelim review.  Tidy things up
         # in these cases.
+        # TODO: make base obsid MetaAttribute in proseco handle this munging.
         if obsid is not None:
             f_obsid = round(float(obsid), 2)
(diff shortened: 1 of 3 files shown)


taldcroft:
cc: @jskrist

taldcroft:
This is pretty straightforward and I'm happy with the tests.  Merging now to get this into master so ORviewer development can see this.

PR #67, opened 2019-02-20
Remove the hack of overloading obsid with roll for roll options

### LINKED ISSUE #66: Roll Options misreporting/miscalculating messages
When an ER (and maybe an OR, but I haven't confirmed it) has roll_options calculated for it, the messages in the roll_options (e.g. acar.roll_options["aca"].messages) do not match the acar.messages.  Specifically, I have seen **Critical** messages become _Warning_ messages:

from the main report:
image
from the roll_options report:
image
Note that the last _Warning_ was a **Critical** message in the main report.

and other times the **Critical** messages seem to disapear:
from the main report:
image
from the roll_options report:
image

taldcroft:
This should fix #66 by removing the hack of overloading obsid with roll.

Cc: @jskrist

This requires https://github.com/sot/proseco/pull/281.

Fixes #66.

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -12,6 +12,7 @@ import pickle
 from itertools import combinations, chain
 
 import matplotlib
+
 matplotlib.use('Agg')
 import matplotlib.pyplot as plt
 from matplotlib.patches import Circle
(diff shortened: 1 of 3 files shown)


jeanconn:
Looks great!  I could reproduce the OR/ER roll issues and confirm that this looks to clean them up on the ERs I looked at in FEB0419 (prepreproducts).  And also it makes more sense.

PR #69, opened 2019-02-20
Remove weakref handling to go along with proseco no-weak branch

### LINKED ISSUE #59: Using roll_level other than 'none' results in error
When I run the method:
`run_aca_review(**kwargs)`
with the input:
`'roll_level': 'none'`
Everything runs as expected and a report is generated for all the acar tables provided, however, when the 'roll_level' is anything other than 'none', I get the following error:

```
<class 'AttributeError'>: 'NoneType' object has no attribute 'fid_set'
 Traceback (most recent call last):
   File "<string>", line 1, in <module>
   File "<string>", line 22, in <module>
   File "[PythonPath]\lib\site-packages\sparkles-4.1-py3.6.egg\sparkles\core.py", line 156, in run_aca_review
     aca.check_catalog()
   File "[PythonPath]\lib\site-packages\sparkles-4.1-py3.6.egg\sparkles\core.py", line 653, in check_catalog
     self.check_acq_p2()
   File "[PythonPath]\lib\site-packages\sparkles-4.1-py3.6.egg\sparkles\core.py", line 771, in check_acq_p2
     P2 = -np.log10(self.acqs.calc_p_safe())
   File "[PythonPath]\lib\site-packages\proseco-4.4-py3.6.egg\proseco\acq.py", line 664, in calc_p_safe
     p_acqs = [acq['probs'].p_acqs(acq['halfw'], man_err) for acq in self]
   File "[PythonPath]\lib\site-packages\proseco-4.4-py3.6.egg\proseco\acq.py", line 664, in <listcomp>
     p_acqs = [acq['probs'].p_acqs(acq['halfw'], man_err) for acq in self]
   File "[PythonPath]\lib\site-packages\proseco-4.4-py3.6.egg\proseco\acq.py", line 1232, in p_acqs
     fid_set = self.acqs().fid_set
 Error executing python statement: 'exec(Path(MX_VAR_reviewScript).open().read())' - 'NoneType' object has no attribute 'fid_set'
```

the script being executed is as follows:

```
full_report_kw_dict = {
"load_name": load_name,
"acas": acas,
"make_html": True,
"outdir": outdir,
"roll_level": roll_level}

path = Path(outdir)
if path.exists():
  shutil.rmtree(path)

from sparkles.core import run_aca_review
run_aca_review(**full_report_kw_dict)
```

taldcroft:
Requires https://github.com/sot/proseco/pull/283

Closes #59

### DIFF
--- a/sparkles/roll_optimize.py
+++ b/sparkles/roll_optimize.py
@@ -353,6 +353,7 @@ class RollOptimizeMixin:
                             del kwargs[key]
 
             kwargs['att'] = q_att_roll
+            kwargs['raise_exc'] = True
 
             aca_rolled = get_aca_catalog(**kwargs)
 
--- a/sparkles/roll_optimize.py
+++ b/sparkles/roll_optimize.py
@@ -343,6 +343,15 @@ class RollOptimizeMixin:
             q_att_roll = calc_aca_from_targ(q_targ_roll, 0, 0)
 
             kwargs = self.call_args.copy()
+
+            # For roll optimization throw away the include/excludes
+            for k1 in ('include', 'exclude'):
+                for k2 in ('ids', 'halfws'):
+                    for k3 in ('acq', 'guide'):
+                        key = f'{k1}_{k2}_{k3}'
+                        if key in kwargs:
+                            del kwargs[key]
+
             kwargs['att'] = q_att_roll
 
             aca_rolled = get_aca_catalog(**kwargs)
--- a/sparkles/roll_optimize.py
+++ b/sparkles/roll_optimize.py
@@ -10,7 +10,7 @@ import numpy as np
 import warnings
 
 from astropy.table import Table, vstack
-from chandra_aca.star_probs import acq_success_prob
+from chandra_aca.star_probs import acq_success_prob, guide_count
 from chandra_aca.transform import (radec_to_yagzag, yagzag_to_pixels,
                                    calc_aca_from_targ, calc_targ_from_aca,
                                    snr_mag_for_t_ccd)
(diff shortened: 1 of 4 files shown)


sparkles/roll_optimize.py
```diff
@@ -353,6 +353,7 @@ def improve_metric(n_stars, P2, n_stars_new, P2_new):
                             del kwargs[key]
 
             kwargs['att'] = q_att_roll
+            kwargs['raise_exc'] = True
```
jeanconn: Remind me why we want this again?

sparkles/tests/test_review.py
```diff
@@ -26,20 +26,19 @@ def test_review_catalog(tmpdir):
...
-         'idx': 2,
-         'text': 'Guide star imposter offset 2.6, limit 2.5 arcsec'},
-        {'category': 'critical', 'text': 'P2: 2.84 less than 3.0 for ER'},
-        {'category': 'critical', 'text':
-         'ER count of 9th (8.9 for -9.9C) mag guide stars 1.57 < 3.0'}]
+        {'text': 'Guide star imposter offset 2.6, limit 2.5 arcsec', 'category': 'warning',
+         'idx': 2},
+        {'text': 'P2: 2.84 less than 3.0 for ER', 'category': 'critical'},
+        {'text': 'ER count of 9th (8.9 for -9.9C) mag guide stars 1.91 < 3.0',
+         'category': 'critical'}]
```
jeanconn: What changed in here?

sparkles/tests/test_review.py
```diff
@@ -117,8 +114,26 @@ def test_probs_weak_reference():
...
+              'man_angle': 131.2011858838081, 'n_acq': 8, 'n_fid': 0, 'n_guide': 8,
+              'sim_offset': 0.0, 'focus_offset': 0.0, 't_ccd_acq': -12.157792574498563,
+              't_ccd_guide': -12.17,
+              'include_ids_acq': np.array(
+                  [8.13042280e+08, 8.13040960e+08, 8.13044168e+08, 8.12911064e+08,
+                   8.12920176e+08, 8.12913936e+08, 8.13043216e+08, 8.13045352e+08]),
+              'include_halfws_acq': np.array(
+                  [160., 160., 160., 160., 160., 160., 120., 60.])}
+
+    aca = get_aca_catalog(**kwargs)
```
jeanconn: I suppose the test does not need to raise_exc=True if we are moving that to the default.

sparkles/roll_optimize.py
taldcroft: Now that it will be the default, not really necessary, but it doesn't hurt.  The need for this was highlighted by the issue yesterday with getting roll options with include_ids set. Debugging this was more difficult because the exception got swallowed and what finally ended up breaking was obscure and unrelated to the root problem.

jeanconn: I just don't understand why we want to raise the exception when this will be called by the MATLAB tools and I don't think the exception is handled in our code.

sparkles/tests/test_review.py
taldcroft: Using chandra_aca.star_probs.guide_count instead of the (previous) local implementation.

jeanconn: Ah, so 1.57 -> 1.91 is the change and the rest is formatting.

sparkles/tests/test_review.py
taldcroft: `acar = aca.get_review_table()` will fail (with an exception) if `aca` is not a well-formed catalog with all the necessary columns.  So this is a valid test with or without the default change.

sparkles/roll_optimize.py
jeanconn: Specifically, on the MATLAB side, we can set raise_exc to False (or whatever @jskrist wants) in get_aca_catalog, but if raise_exc is explicitly set to True in the roll catalogs fetching with no controlling option when called from MATLAB, wouldn't there need to be new code to handle exceptions in the MATLAB Python?

taldcroft: Have you checked the sparkles tidy?  https://github.com/sot/sparkles/pull/71

jeanconn: Ah, you mean the new code to add an option to control this is in the next PR?  I can deal with that.

PR #71, opened 2019-02-21
Tidy code and API

taldcroft:
This changes the API to be more consistent:

- Use `acar` or `acars` for `ACAReviewTable` objects
  - Impacts key in `roll_options` dicts (`'aca'` => `'acar'`)
  - Module-level `run_aca_review` kwarg `acas` => `acars`
- Module-level `run_aca_review` kwarg `outdir` => `report_dir`
- Command line `--outdir` => `--report-dir`
- `ACAReviewTable.run_aca_review`:
  - Now takes a `make_html` boolean input (default=False)
  - Default for `report_dir` is `'.'`
- Add a `raise_exc` arg to `run_aca_review` and `ACAReviewTable.run_aca_review`. 
  -  **ORviewer should call with `raise_exc=False` and check the return value `exc`.  This will be either `None` (all OK) or the traceback in string form (not OK).**

@jskrist

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -376,7 +376,8 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
             del self['idx']
             self.rename_column('idx_temp', 'idx')
 
-    def run_aca_review(self, *, report_dir=None, report_level='none', roll_level='none'):
+    def run_aca_review(self, *, make_html=False, report_dir='.', report_level='none',
+                       roll_level='none'):
         """Do aca review based for this catalog
 
         The ``report_level`` arg specifies the message category at which the full
@@ -386,15 +387,14 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         default is "none", meaning no reports are generated.  A final option is
         "all" which generates a report for every obsid.
 
-        :param report_dir: output directory for report
+        :param make_html: make HTML report (default=False)
+        :param report_dir: output directory for report (default='.')
         :param report_level: report level threshold for generating acq and guide report
         :param roll_level: level threshold for suggesting alternate rolls
 
         :returns: ACAReviewTable object
         """
         acas = [self]
-
-        make_html = (report_dir is not None)
 
         # Do aca review checks and update acas[0] in place
         run_aca_review(load_name=f'Obsid {self.obsid}', acas=acas, make_html=make_html,
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -46,9 +46,9 @@ def main(sys_args=None):
     parser.add_argument('load_name',
                         type=str,
                         help='Load name (e.g. JAN2119A) or full file name')
-    parser.add_argument('--outdir',
+    parser.add_argument('--report-dir',
                         type=str,
-                        help='Output directory (default=<load name>')
+                        help='Report output directory (default=<load name>')
     parser.add_argument('--report-level',
                         type=str,
                         default='none',
@@ -70,14 +70,12 @@ def main(sys_args=None):
                         help='Run quietly')
     args = parser.parse_args(sys_args)
 
-    run_aca_review(args.load_name, outdir=args.outdir,
-                   loud=(not args.quiet), report_level=args.report_level,
-                   roll_level=args.roll_level, obsids=args.obsid)
+    run_aca_review(args.load_name, report_dir=args.report_dir, report_level=args.report_level,
+                   roll_level=args.roll_level, loud=(not args.quiet), obsids=args.obsid)
 
 
-def run_aca_review(load_name=None, *, acas=None, make_html=True, outdir=None,
-                   report_level='none', roll_level='none', loud=False,
-                   obsids=None):
+def run_aca_review(load_name=None, *, acas=None, make_html=True, report_dir=None,
+                   report_level='none', roll_level='none', loud=False, obsids=None):
     """Do ACA load review based on proseco pickle file from ORviewer.
 
     The ``load_name`` specifies the pickle file from which the ``ACATable``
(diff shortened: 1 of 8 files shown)


sparkles/core.py
```diff
@@ -106,39 +107,58 @@ def run_aca_review(load_name=None, *, acas=None, make_html=True, outdir=None,
...
-            outdir = re.sub(r'(_proseco)?.pkl', '', load_name) + '_sparkles'
-        outdir = Path(outdir)
-        outdir.mkdir(parents=True, exist_ok=True)
+            report_dir = re.sub(r'(_proseco)?.pkl', '', load_name) + '_sparkles'
+        report_dir = Path(report_dir)
+        report_dir.mkdir(parents=True, exist_ok=True)
 
     # Do the sparkles review for all the catalogs
-    for aca in acas:
+    for aca in acars:
```
jskrist: for consistency:

`for acar in acars:`

sparkles/core.py
```diff
@@ -106,39 +107,58 @@ def run_aca_review(load_name=None, *, acas=None, make_html=True, outdir=None,
...
         if not isinstance(aca, ACAReviewTable):
```
jskrist: throughout this function:
aca -> acar

sparkles/core.py
```diff
@@ -178,23 +198,25 @@ def run_aca_review(load_name=None, *, acas=None, make_html=True, outdir=None,
...
+        from . import test
+        sparkles_version = test(get_version=True)
         context = {}
         context['load_name'] = load_name.upper()
         context['proseco_version'] = PROSECO_VERSION
-        context['aca_preview_version'] = ACA_PREVIEW_VERSION
-        context['acas'] = acas
-        context['summary_text'] = get_summary_text(acas)
+        context['sparkles_version'] = sparkles_version
+        context['acas'] = acars
```
jskrist: I don't know if this can be updated to use the key 'acars' due to the way the template is setup that this variable is being used for, but if it can handle an input of 'acars', then this should be updated, otherwise, a comment here noting the use in a template which treats these inputs as ACATables, would be good.

jeanconn:
Ah, And I hadn't even seen that this PR was against no-weak yet.  Context is key.

jeanconn:
I'm still a little confused about how an exception caused by something below the forced raise_exc in the roll ranges

https://github.com/sot/sparkles/blob/tidy-spark-joy/sparkles/roll_optimize.py#L321

ends up not propagating back up into ORViewer.

taldcroft:
@jeanconn - the only visibility ORviewer has into this code is via the two `run_aca_review` calls (the function and the method).  Both of those are protected with `raise_exc`, so with `raise_exc=False` there is no way that an exception near https://github.com/sot/sparkles/blob/tidy-spark-joy/sparkles/roll_optimize.py#L321 can filter up to ORviewer.

At least that's how it works in my head, have I made a mistake?

taldcroft:
@jeanconn - are you thinking that `get_roll_options` is called from ORviewer?  That is not part of the "public API" if you will.

jskrist:
Currently, ORViewer only uses `acar.run_aca_review()`, and then `acar.roll_options` to access the roll_options.

jeanconn:
How are there roll options in acar.roll_options if get_roll_options is not called?

taldcroft:
They become available in the process of running the review.

jeanconn:
I just meant that I thought get_roll_options was called as part of run_aca_review.  But get_roll_options sets raise_exc to True on all the catalogs it is getting?  I'll construct an issue with a catalog at a roll and make sure I understand that this is right for raise_exc = False.

taldcroft:
@jeanconn - I just hacked the code to put a `1 / 0` at the point in question and it behaves as advertised.  If you want to verify yourself then do that and use the `test_roll_options_with_include_ids` kwargs and code, with or without `raise_exc=False`.  I'm not sure why you are getting confused, it's just a simple exception catching at the edge of the process.

jeanconn:
Where is the exception caught?

jeanconn:
Really the reason I'm confused, despite it being "simple exception handling", is that I'm trying to keep progress going on this while not actually having time to look at it :smile:

taldcroft:
This is how it will appear to ORviewer:
```
In [13]: out = acar.run_aca_review(roll_level='all', raise_exc=False)

In [14]: print(out)
Traceback (most recent call last):
  File "/Users/aldcroft/git/sparkles/sparkles/core.py", line 125, in run_aca_review
    roll_level=roll_level, loud=loud, obsids=obsids)
  File "/Users/aldcroft/git/sparkles/sparkles/core.py", line 175, in _run_aca_review
    aca.get_roll_options()  # sets roll_options, roll_info attributes
  File "/Users/aldcroft/git/sparkles/sparkles/roll_optimize.py", line 322, in get_roll_options
    1 / 0
ZeroDivisionError: division by zero
```

sparkles/core.py
```diff
@@ -106,39 +107,58 @@ def run_aca_review(load_name=None, *, acas=None, make_html=True, outdir=None,
...
+    :returns: exception message: str or None
 
     """
-    if acas is None:
-        acas = get_acas_from_pickle(load_name, loud)
+    try:
+        _run_aca_review(load_name=load_name, acars=acars, make_html=make_html,
+                        report_dir=report_dir, report_level=report_level,
+                        roll_level=roll_level, loud=loud, obsids=obsids)
+    except Exception:
```
taldcroft: @jeanconn - the exception is caught here.

jeanconn: Thanks!  That's all I needed.  So for interface, @jskrist just needs to change both the `get_aca_catalog` call to add `raise_exc=False` and the `acar.run_aca_review(raise_exc=False)` to get full protection and not have confusing stuff at console.

jskrist: Done and done.  Now if I could only get roll_option messages without generating an html file: #73

taldcroft:
@jskrist - I'll fix the consistency changes you noted.  For now I'm going to merge this just to get a fresh baseline for subsequent (single-issue) patches.  Stand-by...

PR #74, opened 2019-02-21
Fix issue in naming of roll option plots

taldcroft:
The roll option plots had previously all wound up with the same name.  This was introduced in the PR that "improved" the roll handling.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,11 +7,11 @@ try:
 except ImportError:
     cmdclass = {}
 
-entry_points = {'console_scripts': ['sparkles=sparkles.preview:main']}
+entry_points = {'console_scripts': ['sparkles=sparkles.core:main']}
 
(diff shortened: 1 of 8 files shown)


jeanconn:
I think this can be re-targeted as a PR against master.

jeanconn:
So now I assume this is overcome by events by #76 and can be closed?

taldcroft:
Not exactly overcome by events, but the commit of relevance here 4e3fcd8 is in #76.

jeanconn:
Right.  I thought OBE was an OK description because that other commit was
merged in master.  Maybe should be Overcome By Commits.

taldcroft:
OBE to me means the situation changed and the fix is no longer necessary.  My usual wording in this case is that a PR is superceded by another.

PR #77, opened 2019-02-21
Populate roll options with messages independent of HTML reporting

### LINKED ISSUE #73: roll_options.messages not created when make_html == False
When I run:

`acar.run_aca_review(**kwargs)`

for an ACAReviewTable object using the kwargs:
`{"report_dir": ".", "roll_level": "all"}`

The messages for the first roll option (which should be the original acar) is empty.  However, if I specify the kwargs:
`{"report_dir": ".", "roll_level": "all", "make_html": True}`

Then the messages are filled.

taldcroft:
This continues the morph of a standalone tool that just made a web page into a real thing with user-accessible objects.  The original design wasn't quite ideal for the latter purpose, but we will get there eventually.  Sooner or later I will restructure this code to be more reasonable (without API changes).  In the mean time this should fix #73.

This does end up re-doing the check_catalog() call in the case of making HTML, but that call is so fast compared to making plots etc that I'm not too concerned, at least for now.

Fixes #73 

cc: @jskrist

### DIFF
--- a/sparkles/tests/test_review.py
+++ b/sparkles/tests/test_review.py
@@ -43,7 +43,15 @@ def test_review_catalog(tmpdir):
 
     assert acar.review_status() == -1
 
+    # Run the review but without making the HTML and ensure review messages
+    # are available on roll options.
+    acar.run_aca_review(roll_level='critical')
+    assert len(acar.roll_options) > 1
+    assert acar.roll_options[0]['acar'].messages == acar.messages
+    assert len(acar.roll_options[1]['acar'].messages) > 0
+
     # Check doing a full review for this obsid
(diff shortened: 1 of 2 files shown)


sparkles/tests/test_review.py
```diff
@@ -82,7 +90,7 @@ def test_review_roll_options(tmpdir):
 
     aca = get_aca_catalog(**kwargs)
     acar = aca.get_review_table()
-    acar.run_aca_review(make_html=True, report_dir=tmpdir, roll_level='critical')
+    acar.run_aca_review(roll_level='critical')
```
taldcroft: With this fix there is no need for this test to be writing HTML.

PR #80, opened 2019-02-22
Remove the line that does nothing now but was needed before the API change

taldcroft:


jeanconn:
Thanks!  And sorry yeah, I was just reading the code to try to infer what was going on instead of setting breakpoints and evaluating values... and this one line just threw me because I didn't realize it was part of the initial hack to get the roll options debugging going.

PR #82, opened 2019-02-22
Set make_starcat_plot to always make/remake the images

### LINKED ISSUE #81: Should probably add something to check to see if plots need to be remade
The MAR0419 second prereview has at least one starcat plot that doesn't match the catalog.  My best guess is that the tool was run more than once on a pickle file of the same name (but with different catalogs) and that the line here:

https://github.com/sot/sparkles/blob/ee9455adc4034932cfdc693db0f4a0afb813bdde/sparkles/core.py#L600

kept the catalog that had actually changed from making a new plot.  If that line is there as an optimization to make sure we don't keep making plots gratuitously, we probably need something (file saved sha of catalog?) to check before deciding if the plot gets remade or not.  I assume this could also be an issue in the view-the-html reporting from ORViewer (will the catalog plots actually get remade after an edit?  That was really the only use case I put forward at the working group).

jeanconn:
Set make_starcat_plot to always make/remake the images

This cuts the previous optimization that skipped making the plot if it already existed for the observation.  While it might exist, it could be stale.

Closes #81

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -597,9 +597,6 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         outfile = self.obsid_dir / plotname
         self.context['catalog_plot'] = outfile.relative_to(self.preview_dir).as_posix()
 
-        if outfile.exists():
-            return
-
         fig = plt.figure(figsize=(6.75, 6.0))
         ax = fig.add_subplot(1, 1, 1)
         self.plot(ax=ax)


jeanconn:
Though I'm not actually sure if this fixes my use case just yet.  Still testing.

jeanconn:
Ah.  I'd gotten confused that the png actually changed names between sparkles versions (so my outdated named file was not overwritten).  Manual test now looks good.

PR #89, opened 2019-02-26
Remove use of test(get_version=True) for versioning

### LINKED ISSUE #88: remove test() calls in release code
https://github.com/sot/sparkles/blob/9f330762c485e088fc5a16ebdecff82d87ada56e/sparkles/core.py#L36

This code causes a dialog to flash in Windows, which is undesirable for the release.

taldcroft:
This passes unit tests.  I confirmed that running on a pickle file from JAN2819 and generating a review report shows the expected version numbers.

Fixes #88.

I'm basing this on 4.1.x.  Once reviewed and merged we can cherry-pick into master.  Traditionally it goes the other way around, but getting this in 4.1.x first is probably the priority.

### DIFF
--- a/sparkles/__init__.py
+++ b/sparkles/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '4.1'
+__version__ = '4.1.1'
 
 from .core import run_aca_review, ACAReviewTable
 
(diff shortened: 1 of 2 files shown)


jeanconn:
And since master and 4.1.x should be the same, we should be OK regardless.

taldcroft:
For the record I just cherry-picked this into master:
```
$ get checkout master
$ git cherry-pick -m 1  68ac502
$ git diff master 4.1.x
  <no diff>
$ git push origin master
```
That generated 4dfc3e6035c4ef6dda8773f6e511535105533f5b.

PR #91, opened 2019-02-26
Fix bug in get_roll_options for small or zero allowed roll deviation

taldcroft:
This problem appeared for a far-forward observation where the allowed roll deviation is exactly 0.0, and the actual scheduled roll ends up being outside the min/max determined by the code.  This caused the index issue because the code requires roll is inside the range.  The fix is just to expand `roll_min` or `roll_max` to be sure `roll` is contained within.

This PR also adds exception handling in the same pattern as for `make_report()`, but does one better by including the full traceback.

To do:
- [ ] Add a test.  I found a way to trigger an exception in get_roll_options() from outside, so will add the test tonight.

For the record these are kwargs that showed the original issue, though the only important ones are attitude and date.
```
{'att': [-0.82389459, -0.1248412, 0.35722113, 0.42190692],
 'date': '2019:073:21:55:30.000',
 'detector': 'ACIS-S',
 'dither_acq': (7.9992, 7.9992),
 'dither_guide': (7.9992, 7.9992),
 'focus_offset': 0.0,
 'man_angle': 122.97035882921071,
 'n_acq': 8,
 'n_fid': 0,
 'n_guide': 8,
 'obsid': 48334.0,
 'sim_offset': 0.0,
 't_ccd_acq': -10.257559323423214,
 't_ccd_guide': -10.25810835536192}
```
And the exception:
```
~/miniconda3/envs/ska3-test/lib/python3.6/site-packages/sparkles/core.py in _run_aca_review(load_name, acars, make_html, report_dir, report_level, roll_level, loud, obsids)
    171 
    172         if roll_level == 'all' or aca.messages >= roll_level:
--> 173             aca.get_roll_options()  # sets roll_options, roll_info attributes
    174 
    175         if make_html:

~/miniconda3/envs/ska3-test/lib/python3.6/site-packages/sparkles/roll_optimize.py in get_roll_options(self)
    285 
    286         cand_idxs = self.get_candidate_better_stars()
--> 287         roll_intervals, self.roll_info = self.get_roll_intervals(cand_idxs)
    288 
    289         q_att = self.att

~/miniconda3/envs/ska3-test/lib/python3.6/site-packages/sparkles/roll_optimize.py in get_roll_intervals(self, cand_idxs, roll_nom, roll_dev, y_off, z_off, d_roll)
    197         # and not roll_nom.
    198         ids_list = get_ids_list(roll_offsets)
--> 199         ids0 = ids_list[len(ro_minus)]
    200 
    201         # Get all unique sets of stars that are in the FOV over the sampled

IndexError: list index out of range
```

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -170,7 +170,13 @@ def _run_aca_review(load_name=None, *, acars=None, make_html=True, report_dir=No
         aca.check_catalog()
 
         if roll_level == 'all' or aca.messages >= roll_level:
-            aca.get_roll_options()  # sets roll_options, roll_info attributes
+            try:
+                aca.get_roll_options()  # sets roll_options, roll_info attributes
+            except Exception: # as err:
+                err = traceback.format_exc()
+                aca.add_message('critical', text=f'Running get_roll_options() failed: \n{err}')
+                aca.roll_options = None
+                aca.roll_info = None
 
         if make_html:
 
@@ -186,8 +192,9 @@ def _run_aca_review(load_name=None, *, acars=None, make_html=True, report_dir=No
             if report_level == 'all' or aca.messages >= report_level:
                 try:
                     aca.make_report()
-                except Exception as err:
-                    aca.add_message('critical', text=f'Running make_report() failed: {err}')
+                except Exception:
+                    err = traceback.format_exc()
+                    aca.add_message('critical', text=f'Running make_report() failed:\n{err}')
 
             if aca.roll_info:
                 aca.make_roll_options_report()
--- a/sparkles/roll_optimize.py
+++ b/sparkles/roll_optimize.py
@@ -178,16 +178,19 @@ class RollOptimizeMixin:
         if roll_dev is None:
             roll_dev = allowed_rolldev(pitch)
 
-        # Ensure roll_nom in range 0 <= roll_nom < 360 to match q_att.roll
+        # Ensure roll_nom in range 0 <= roll_nom < 360 to match q_att.roll.
+        # Also ensure that roll_min < roll < roll_max.  It can happen that the
+        # ORviewer scheduled roll is outside the allowed_rolldev() range.  For
+        # far-forward sun, allowed_rolldev() = 0.0.
+        roll = q_att.roll
         roll_nom = roll_nom % 360.0
-        roll_min = roll_nom - roll_dev
-        roll_max = roll_nom + roll_dev
+        roll_min = min(roll_nom - roll_dev, roll - 0.1)
+        roll_max = max(roll_nom + roll_dev, roll + 0.1)
 
         # Get roll offsets spanning roll_min:roll_max with padding.  Padding
         # ensures that if a candidate is best at or beyond the extreme of
         # allowed roll then make sure the sampled rolls go out far enough so
         # that the mean of the roll_offset boundaries will get to the edge.
-        roll = q_att.roll
         ro_minus = np.arange(0, roll_min - roll_dev - roll, -d_roll)[1:][::-1]
         ro_plus = np.arange(0, roll_max + roll_dev - roll, d_roll)
         roll_offsets = np.concatenate([ro_minus, ro_plus])
--- a/sparkles/__init__.py
+++ b/sparkles/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '4.1.1'
+__version__ = '4.1.2'
 
 from .core import run_aca_review, ACAReviewTable
 


sparkles/core.py
```diff
@@ -170,7 +170,13 @@ def _run_aca_review(load_name=None, *, acars=None, make_html=True, report_dir=No
         aca.check_catalog()
 
         if roll_level == 'all' or aca.messages >= roll_level:
-            aca.get_roll_options()  # sets roll_options, roll_info attributes
+            try:
+                aca.get_roll_options()  # sets roll_options, roll_info attributes
+            except Exception: # as err:
+                err = traceback.format_exc()
+                aca.add_message('critical', text=f'Running get_roll_options() failed: \n{err}')
+                aca.roll_options = None
```
jeanconn: This looks great to me but I'm not clear on if it is an API change (were these None when this failed previously or did the attributes just not exist?).

jeanconn:
So in the bug, had at least the roll_nom values ended up in the checked rolls (but not the scheduled roll)?

sparkles/roll_optimize.py
```diff
@@ -178,16 +178,19 @@ def get_ids_list(roll_offsets):
...
+        # Ensure roll_nom in range 0 <= roll_nom < 360 to match q_att.roll.
+        # Also ensure that roll_min < roll < roll_max.  It can happen that the
+        # ORviewer scheduled roll is outside the allowed_rolldev() range.  For
+        # far-forward sun, allowed_rolldev() = 0.0.
+        roll = q_att.roll
         roll_nom = roll_nom % 360.0
-        roll_min = roll_nom - roll_dev
-        roll_max = roll_nom + roll_dev
+        roll_min = min(roll_nom - roll_dev, roll - 0.1)
+        roll_max = max(roll_nom + roll_dev, roll + 0.1)
```
jeanconn: And these look good to me but I'm not sure about the 0.1 values.

jeanconn: My concern with the 0.1 values was just that for d_roll = 0.25, I didn't know if you could get into a situation with, say at the positive case, roll - roll_nom = 0.16, and then end up with the range for a zero roll_dev containing roll_nom, roll, *and* an extra candidate roll.  Given that the point of the tool is to provide rolls to check that are already advertised as experimental and perhaps not allowed, this seems fine (especially for the weird case where roll_dev is 0 and roll is a good chunk (> 0.15) from roll_nom etc).

sparkles/core.py
taldcroft: The `None` values correspond to the defaults in `__init__`, so these are consistent with no attempt to get roll options.  So the API is to rely on messages to see a failure.  Hopefully this never actually gets used, but I think that seeing a critical message at this level will get the required attention.

taldcroft:
> So in the bug, had at least the roll_nom values ended up in the checked rolls (but not the scheduled roll)?

With the bug in place it just crashed and there were no checked rolls.  The list of rolls that were going to be checked included only the scheduled roll (roll offset = 0).

sparkles/roll_optimize.py
taldcroft: The case that actually caused the problem had roll - roll_nom of around 0.34.  As you say, there is really no concern with extra candidate rolls, in fact the fundamental design is always to sample rolls that are outside of the allowed range.  In fact another thought I had for fixing things was basically:
```
        padding = roll_dev + d_roll * 2  # Make sure there are at least a few samples
        ro_minus = np.arange(0, roll_min - padding - roll, -d_roll)[1:][::-1]
        ro_plus = np.arange(0, roll_max + padding - roll, d_roll)
```
But at the end this won't really help because the allowed options will be clipped by roll_min and roll_max.  And it *is important* to make sure that `roll_min <= roll <= roll_max`, so the simpler fix of just opening up roll_min and roll_max seemed better.

jeanconn: Regarding "And it _is important_ to make sure that `roll_min <= roll <= roll_max`, so the simpler fix of just opening up roll_min and roll_max seemed better.",  right, I was trying to wrap my head around any other implicit requirements.

jskrist: I know this is still a beta feature, but I am unclear about which piece of data is "real" or "more important".  If the `roll_min`/`roll_max` can be (arbitrarily) expanded, then is the reported `roll` the "real" thing that needs to be safeguarded?

Another way to put this:

It seems, from the description of the issue, that this could be fixed by either expanding the `roll_min` - `roll_max` range to always include the `roll`, **or** by moving the `roll` to always be inside the `roll_min` - `roll_max` range.

Why is expanding the range the "right" choice?

jeanconn: If I'm understanding correctly, in this case the as-scheduled roll (aka "roll") was outside the tool's calculated range.  So I was with Tom in that if the mission planners supply knowledge somewhat counter to the tool, it is relatively reasonable to just go along with that instead of returning a "your observation was outside the roll range we've calculated (apparently badly)".

jskrist: ah, that makes sense.  I was making the assumption that the roll is usually being chosen by proseco or based on some estimation of what the roll should be, not coming from the mission planner.  I agree that if the roll is provided by the mission planner, it would likely be best to not modify it.

jeanconn: And I still said it badly, but I think we're on the same page 👍

taldcroft: Agreed with what @jeanconn said.

sparkles/core.py
taldcroft: @jskrist - this has the potential to put in a multiline message with a `\n` in it.  It works fine in HTML.  I assume within Windows MATLAB that this won't *break* anything, and at worst just make something that isn't formatted perfectly.  I could probably make that `\n` into `os.linesep()`, but I don't know if that will help out.

BTW the only way I can think of to test this in ORviewer is to hack the code (put in a `x = 1/0` statement) and install that.   Right now I don't know any way to break the code from the outside. 

One way which I plan for testing is to make an `acar` object and then do `del acar.stars['ra']` and then do the review with roll options.   But I don't think that helps in ORviewer integration testing.

jskrist: @taldcroft - thanks for the heads up.  I can force the situation in ORViewer in a similar way (run a schedule update and check for stars [with a breakpoint between where it creates the ACAReviewTables and where it runs the report with roll_options], then manually modify the acar object and then do the review with roll options.  It is not easy and won't be able to be easily automated, but I can do it manually for now.

Also, I _think_ the newline character ("\n") will be interpreted correctly, but I will double check, as I might be escaping any backslashes that I don't expect (for safety).

PR #93, opened 2019-02-27
Roll around the correct aimpoint, redux

taldcroft:
Sparkles is sparking no joy right now...

79aff0615ded0 only did half the job.  This finishes it up.  For validation work this is based off of pickle-roll-options branch, see ad8c820.

### DIFF
--- a/sparkles/roll_optimize.py
+++ b/sparkles/roll_optimize.py
@@ -290,7 +290,7 @@ class RollOptimizeMixin:
         roll_intervals, self.roll_info = self.get_roll_intervals(cand_idxs)
 
         q_att = self.att
-        q_targ = calc_targ_from_aca(q_att, 0, 0)
+        q_targ = self._calc_targ_from_aca(q_att, 0, 0)
 
         # Special case, first roll option is self but with obsid set to roll
         acar = deepcopy(self)
@@ -309,7 +309,7 @@ class RollOptimizeMixin:
         for roll_interval in roll_intervals:
             roll = roll_interval['roll']
             q_targ_roll = Quat([q_targ.ra, q_targ.dec, roll])
-            q_att_roll = calc_aca_from_targ(q_targ_roll, 0, 0)
+            q_att_roll = self._calc_aca_from_targ(q_targ_roll, 0, 0)
 
             kwargs = self.call_args.copy()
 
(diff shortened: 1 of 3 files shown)


sparkles/roll_optimize.py
```diff
@@ -309,7 +309,7 @@ def improve_metric(n_stars, P2, n_stars_new, P2_new):
         for roll_interval in roll_intervals:
             roll = roll_interval['roll']
             q_targ_roll = Quat([q_targ.ra, q_targ.dec, roll])
-            q_att_roll = calc_aca_from_targ(q_targ_roll, 0, 0)
+            q_att_roll = self._calc_aca_from_targ(q_targ_roll, 0, 0)
```
jeanconn: Looks great.  Though I'm assuming if this wasn't noticed that this could use a quick test.  I could help with that.

taldcroft: Sure, thanks!

jeanconn: I think this would also be more readable if those were q_att_rolled and q_targ_rolled, but that might just be me.

jeanconn:
Also, I haven't done the math, but does the fix in ad8c820 mean that the fix in 91267be is still required?

sparkles/roll_optimize.py
taldcroft: Done.

taldcroft:
> Also, I haven't done the math, but does the fix in [ad8c820](https://github.com/sot/sparkles/commit/ad8c8200dafec1a9936420b9d72078fc7fe79363) mean that the fix in [91267be](https://github.com/sot/sparkles/commit/91267bee86609f321e785aa93d7887933b8946a0) is still required?

The 91267be fix is still necessary.  In the case of a 0.0 allowed roll deviation (which was the case for the original bug), it will always be true that the ORviewer supplied `roll` is not within the `roll_min` to `roll_max` range, because that range had zero width (before the fix), and ORviewer is using a different algorithm for allowed roll range.

taldcroft:
@jeanconn - can you take care of getting the open/approved PRs merged?  I don't care if you do the individual ones or just this PR which is the glop of them together.  I'm good with all.

taldcroft:
And I'm going to start work on the target roll stuff based on the roll-opt-targ-bug branch, so it'll be nice to get this all into master.

PR #94, opened 2019-02-27
Add some roll-option related tests

jeanconn:
Add some roll_option related tests.

I ended up not being able to come up with a trivial test for https://github.com/sot/sparkles/commit/ad8c8200dafec1a9936420b9d72078fc7fe79363 as I think we'd need to construct a case with very different roll_options that would show up just with the targ vs aca offset of ODB_SI_ALIGN applied.  I did not try to construct or find such a case.  Instead I just added a regression test on get_roll_intervals() for now.

### DIFF
--- a/sparkles/tests/test_review.py
+++ b/sparkles/tests/test_review.py
@@ -187,6 +187,14 @@ def test_run_aca_review_function():
 
 
 def test_roll_outside_range():
+    """
+    Run a test on obsid 48334 from ~MAR1119 that is at a pitch that has 0 roll_dev
+    and is at a roll that is not exactly nominal roll for the attitude and date.
+    The 'att' ends up with roll outside of the internally-computed roll_min / roll_max
+    which caused indexing excption in the roll-options code.  Fixed by PR #91 which
+    includes code  to expand the roll_min / roll_max range to always include the roll
+    of the originally supplied attitude.
+    """
     kw = {'att': [-0.82389459, -0.1248412, 0.35722113, 0.42190692],
           'date': '2019:073:21:55:30.000',
           'detector': 'ACIS-S',
@@ -213,21 +221,20 @@ def test_calc_targ_from_aca():
     Confirm _calc_targ_from_aca seems to do the right thing based on obsid
     This does a bit too much processing for what should be a lightweight test.
     """
-    obs_kwargs = KWARGS_48464
-    aca_er = get_aca_catalog(**obs_kwargs)
-    acar_er = aca_er.get_review_table()
-    q_targ = acar_er._calc_targ_from_aca(acar_er.att, 0, 0)
-    # Cheat and use the string reprs for comparison to 8 decimals
-    assert Quat(obs_kwargs['att']).__repr__() == q_targ.__repr__()
+    # Testing an ER where we expect the targ quaternion to be the same as the ACA
+    # quaternion.
+    aca = get_aca_catalog(**KWARGS_48464)
+    acar = aca.get_review_table()
+    q_targ = acar._calc_targ_from_aca(acar.att, 0, 0)
+    assert q_targ is acar.att
 
-    or_obs_kwargs = obs_kwargs.copy()
-    or_obs_kwargs['obsid'] = 1
-    aca_or = get_aca_catalog(**or_obs_kwargs)
-    acar_or = aca_or.get_review_table()
-    q_targ = acar_or._calc_targ_from_aca(acar_or.att, 0, 0)
-    # Assert that the the OR version *doesn't* match and is off by about 70 arcsecs yaw
-    assert Quat(obs_kwargs['att']).__repr__() != q_targ.__repr__()
-    assert abs(Quat(obs_kwargs['att']).dq(q_targ).yaw * 3600) - 69.59 < .01
+    # Here we change the review object to represent an OR (by changing is_OR) and
+    # confirm the targ quaternion is different from the ACA quaternion
+    acar._is_OR = True
+    q_targ = acar._calc_targ_from_aca(acar.att, 0, 0)
+    # Specifically the targ quaternion should be off by ODB_SI_ALIGN which is about 70
+    # arcsecs in yaw
+    assert np.isclose(acar.att.dq(q_targ).yaw * 3600, 69.59, atol=0.01, rtol=0)
 
 
 def test_get_roll_intervals():
@@ -262,8 +269,8 @@ def test_get_roll_intervals():
         acar_or.get_candidate_better_stars(),
         roll_dev=roll_dev, y_off=y_off, z_off=z_off)
 
-    assert Quat(obs_kwargs['att']).roll <= er_info['roll_max']
-    assert Quat(obs_kwargs['att']).roll >= er_info['roll_min']
+    assert acar_er.att.roll <= er_info['roll_max']
+    assert acar_er.att.roll >= er_info['roll_min']
 
     # The roll ranges in ACA rolls should be the same for both the ER and the OR version
     assert or_info == er_info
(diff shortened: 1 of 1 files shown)


jeanconn:
I'm not thrilled with these.  This PR is set against roll-opt-targ-bug if you want any or none of them to go in @taldcroft .

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,113 @@ def test_run_aca_review_function():
         {'text': 'P2: 2.84 less than 3.0 for ER', 'category': 'critical'},
         {'text': 'ER count of 9th (8.9 for -9.9C) mag guide stars 1.91 < 3.0',
          'category': 'critical'}]
+
+
+def test_roll_outside_range():
+    kw = {'att': [-0.82389459, -0.1248412, 0.35722113, 0.42190692],
```
taldcroft: OK, just needs a docstring with reference to the PR #91 this is testing and an explanation of why this had previously failed.  (I.e just say in a sentence that this attitude ends up with roll outside of the internally-computed roll_min / roll_max.).

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,113 @@ def test_run_aca_review_function():
...
+    """
+    Confirm _calc_targ_from_aca seems to do the right thing based on obsid
+    This does a bit too much processing for what should be a lightweight test.
+    """
+    obs_kwargs = KWARGS_48464
+    aca_er = get_aca_catalog(**obs_kwargs)
+    acar_er = aca_er.get_review_table()
+    q_targ = acar_er._calc_targ_from_aca(acar_er.att, 0, 0)
+    # Cheat and use the string reprs for comparison to 8 decimals
+    assert Quat(obs_kwargs['att']).__repr__() == q_targ.__repr__()
```
taldcroft: You should be able to do this one as `assert q_targ is acar_er.att`.

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,113 @@ def test_run_aca_review_function():
...
+
+    or_obs_kwargs = obs_kwargs.copy()
+    or_obs_kwargs['obsid'] = 1
+    aca_or = get_aca_catalog(**or_obs_kwargs)
+    acar_or = aca_or.get_review_table()
+    q_targ = acar_or._calc_targ_from_aca(acar_or.att, 0, 0)
+    # Assert that the the OR version *doesn't* match and is off by about 70 arcsecs yaw
+    assert Quat(obs_kwargs['att']).__repr__() != q_targ.__repr__()
+    assert abs(Quat(obs_kwargs['att']).dq(q_targ).yaw * 3600) - 69.59 < .01
```
taldcroft: 0.01

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,113 @@ def test_run_aca_review_function():
...
+    # Cheat and use the string reprs for comparison to 8 decimals
+    assert Quat(obs_kwargs['att']).__repr__() == q_targ.__repr__()
+
+    or_obs_kwargs = obs_kwargs.copy()
+    or_obs_kwargs['obsid'] = 1
+    aca_or = get_aca_catalog(**or_obs_kwargs)
+    acar_or = aca_or.get_review_table()
+    q_targ = acar_or._calc_targ_from_aca(acar_or.att, 0, 0)
+    # Assert that the the OR version *doesn't* match and is off by about 70 arcsecs yaw
+    assert Quat(obs_kwargs['att']).__repr__() != q_targ.__repr__()
```
taldcroft: This test is not really adding anything.  We can stipulate that if the second passes the first will always pass.

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,113 @@ def test_run_aca_review_function():
...
+    assert Quat(kw['att']).roll <= acar.roll_info['roll_max']
+    assert Quat(kw['att']).roll >= acar.roll_info['roll_min']
+
+
+def test_calc_targ_from_aca():
+    """
+    Confirm _calc_targ_from_aca seems to do the right thing based on obsid
+    This does a bit too much processing for what should be a lightweight test.
+    """
+    obs_kwargs = KWARGS_48464
```
taldcroft: No need for this one-off.

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,113 @@ def test_run_aca_review_function():
...
+    aca_er = get_aca_catalog(**obs_kwargs)
```
taldcroft: Life is easier / simpler to read in tests like this if you just re-use the same variable names all the time (`aca`, `acar`, `kwargs` if need be) but add useful comments like `Testing an ER where we expect the targ quaternion to be the same as the ACA quaternion`.

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,113 @@ def test_run_aca_review_function():
...
+    acar_er = aca_er.get_review_table()
+    q_targ = acar_er._calc_targ_from_aca(acar_er.att, 0, 0)
+    # Cheat and use the string reprs for comparison to 8 decimals
+    assert Quat(obs_kwargs['att']).__repr__() == q_targ.__repr__()
+
+    or_obs_kwargs = obs_kwargs.copy()
```
taldcroft: For testing you can use the back door and make this simpler (and easier to understand later):
```
# Now treat this as an OR by forcing the object to believe it is so,
# and confirm the expected HRMA to ACA aimpoint offset of about 70 arcsec.
acar._is_OR = True
q_targ = acar._calc_targ_from_aca(acar.att, 0, 0)
assert np.isclose(acar.att.dq(q_targ).yaw * 3600, 69.59, atol=0.01, rtol=0)
```

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,113 @@ def test_run_aca_review_function():
...
+
+    er_roll_intervs, er_info = acar_er.get_roll_intervals(
+        acar_er.get_candidate_better_stars(),
+        roll_dev=roll_dev, y_off=y_off, z_off=z_off)
+
+    or_roll_intervs, or_info = acar_or.get_roll_intervals(
+        acar_or.get_candidate_better_stars(),
+        roll_dev=roll_dev, y_off=y_off, z_off=z_off)
+
+    assert Quat(obs_kwargs['att']).roll <= er_info['roll_max']
```
taldcroft: Use `acar_er.att` which is guaranteed to be a Quat.

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,113 @@ def test_run_aca_review_function():
...
+    assert abs(Quat(obs_kwargs['att']).dq(q_targ).yaw * 3600) - 69.59 < .01
+
+
+def test_get_roll_intervals():
+    """
+    Test that the behavior of get_roll_intervals is different for ORs and ERs with
+    regard to use of offsets.  They are set to arbitrary large values in the test.
+    """
+
+    # This uses the catalog at KWARGS_48464, but would really be better as a fully
```
taldcroft: The roll options code doesn't yet work with synthetic catalogs.  That would be a good thing.

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,113 @@ def test_run_aca_review_function():
...
+    # The roll ranges in ACA rolls should be the same for both the ER and the OR version
+    assert or_info == er_info
+
+    # Up to this point this is really a weak functional test.  The following asserts
+    # are more regression tests for the attitude at obsid 48464
+    or_rolls = [interv['roll'] for interv in or_roll_intervs]
+    er_rolls = [interv['roll'] for interv in er_roll_intervs]
+    assert or_rolls != er_rolls
+
+    assert or_roll_intervs == [{'roll': 281.63739755173594,
```
taldcroft: These exact comparisons are asking for trouble.  Define `exp_intervs` as what you've written and then:
```
for interv, exp_interv in zip(or_roll_intervs, exp_intervs):
    assert interv.keys() == exp_interv.keys()
    for key in interv.keys():
        if key.startswith('roll'):
            assert np.isclose(..., atol=1e-6, rtol=0)
        else:
            assert .. a == b
```

sparkles/tests/test_review.py
jeanconn: I was mostly thinking about making it easier to swap in a different choice, but sure.

sparkles/tests/test_review.py
jeanconn: Good point.  I was comparing the ERs to the ORs when I started, but forgot to clean this up when I removed that.

jeanconn:
And I still don't think these tests will be great, but working around in here directly did give me a much better understanding of the roll options code, so that was a good use of time either way.

sparkles/tests/test_review.py
jeanconn: I wasn't sure if that was cheating, but OK.

sparkles/tests/test_review.py
jeanconn: I made a quick method to do this to not repeat the loop for the OR and ER cases.

sparkles/tests/test_review.py
```diff
@@ -182,3 +184,136 @@ def test_run_aca_review_function():
...
+          'n_guide': 8,
+          'obsid': 48334.0,
+          'sim_offset': 0.0,
+          't_ccd_acq': -10.257559323423214,
+          't_ccd_guide': -10.25810835536192}
+    aca = get_aca_catalog(**kw)
+    acar = aca.get_review_table()
+    acar.get_roll_options()
+    assert Quat(kw['att']).roll <= acar.roll_info['roll_max']
+    assert Quat(kw['att']).roll >= acar.roll_info['roll_min']
```
jeanconn: I left this one not using acar.att.roll so that we are explicitly testing what we care about... that the requested roll in the kwargs is in the returned roll range in roll_info.

PR #97, opened 2019-02-28
Use target attitude consistently in roll optimization

### LINKED ISSUE #95: reported "roll" for roll options is different from the acar.att.roll
I might not fully understand this behavior.  This is what I am seeing.  When an html report is generated for an observation, if it has roll_options, a separate roll_options report page is generated.  The rolls reported on this page are different from what I am reporting in ORViewer.  Upon investigation of this discrepancy, I found that:

`acar.roll_options[1]["acar"].att.roll ~= acar.roll_options[1]["roll"]`

This seems to be caused by `acar.roll_options[1]["roll"]` being calculated using `(interval['x_start'] + interval['x_stop']) / 2`.  Is this expected?  Which is the "correct" roll?

taldcroft:
This makes the changes I proposed in #95.  The code changes are actually less than they look because I took the opportunity to make some variable naming more consistent.

But the very important one is that "roll" in the roll info output is **always** the target attitude roll.   For ERs that is the same as the ACA attitude. That's basically what the mission planners want to see.

So by design now there will be a diff between "roll" in the roll options and `acar.roll`, but the roll option roll will match `acar.att_targ.roll`.  Likewise `roll_min` and `roll_max` refer to target attitude roll.

Fixes #95 

cc: @jskrist

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -6,6 +6,7 @@ Preliminary review of ACA catalogs selected by proseco.
 """
 import gzip
 import io
+import os
 import re
 import traceback
 from pathlib import Path
@@ -622,15 +623,20 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         """
         P2 = -np.log10(self.acqs.calc_p_safe())
         att = self.att
+        att_targ = self.att_targ
         self._base_repr_()  # Hack to set default ``format`` for cols as needed
         catalog = '\n'.join(self.pformat(max_width=-1, max_lines=-1))
         self.acq_count = np.sum(self.acqs['p_acq'])
+
+        att_string = f'ACA RA, Dec, Roll (deg): {att.ra:.6f} {att.dec:.5f} {att.roll:.5f}'
+        if self.is_OR:
+            att_string += f'  [Target: {att_targ.ra:.6f} {att_targ.dec:.5f} {att_targ.roll:.5f}]'
 
         message_text = self.get_formatted_messages()
 
         text_pre = f"""\
 {self.detector} SIM-Z offset: {self.sim_offset}
-RA, Dec, Roll (deg): {att.ra:.6f} {att.dec:.5f} {att.roll:.5f}
+{att_string}
 Dither acq: Y_amp= {self.dither_acq.y:.1f}  Z_amp={self.dither_acq.z:.1f}
 Dither gui: Y_amp= {self.dither_guide.y:.1f}  Z_amp={self.dither_guide.z:.1f}
 Maneuver Angle: {self.man_angle:.2f}
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -457,6 +457,12 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         return status
 
     @property
+    def att_targ(self):
+        if not hasattr(self, '_att_targ'):
+            self._att_targ = self._calc_targ_from_aca(self.att, 0, 0)
+        return self._att_targ
+
+    @property
     def report_id(self):
         return round(self.att.roll, 2) if self.is_roll_option else self.obsid
 
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -535,6 +535,8 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
 
         for opt in opts:
             del opt['acar']
+            for name in ('add_ids', 'drop_ids'):
+                opt[name] = ' '.join(str(id_) for id_ in opt[name]) if opt[name] else '--'
 
         opts_table = Table(opts, names=['roll', 'P2', 'n_stars', 'improvement',
                                         'roll_min', 'roll_max', 'add_ids', 'drop_ids'])
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -650,9 +650,9 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         catalog = '\n'.join(self.pformat(max_width=-1, max_lines=-1))
         self.acq_count = np.sum(self.acqs['p_acq'])
 
-        att_string = f'ACA RA, Dec, Roll (deg): {att.ra:.6f} {att.dec:.5f} {att.roll:.5f}'
+        att_string = f'ACA RA, Dec, Roll (deg): {att.ra:.5f} {att.dec:.5f} {att.roll:.5f}'
         if self.is_OR:
-            att_string += f'  [Target: {att_targ.ra:.6f} {att_targ.dec:.5f} {att_targ.roll:.5f}]'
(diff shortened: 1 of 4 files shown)


taldcroft:
This passes unit tests but I haven't tried digging in.  I need to step away for a bit, but maybe @jeanconn will have a look and try to break it with an 89.9 deg declination case?  :smile:

sparkles/core.py
```diff
@@ -456,6 +456,12 @@ def review_status(self):
 
         return status
 
+    @property
+    def att_targ(self):
+        if not hasattr(self, '_att_targ'):
+            self._att_targ = self._calc_targ_from_aca(self.att, 0, 0)
```
jeanconn: This looks good (computing the _att_targ as needed) though I suppose if we get the y_off z_off values there will be more places to scrub.  Also, not sure why _calc_targ_from_aca requires args at all, but that's not an issue for this PR.

taldcroft: Agreed those should be optional kwargs.  When we get offsets in the code here it will be an easy scrub.

jeanconn:
I can't make it error out with an 89.99 dec (or -89.99) case, but not sure what to add to the tests.

taldcroft:
@jeanconn - I'm working on a functional test now, so you don't need to do any more on this for now (in terms of testing).

taldcroft:
See:

https://icxc.cfa.harvard.edu/aspect/tmp/obs18000/
https://icxc.cfa.harvard.edu/aspect/tmp/obs48000/

These are the outputs for an ER and OR at a dec=89.9, where the difference between ACA and target frames is significant.  If you want to play along at home:
```
from proseco.tests.test_common import mod_std_info
from Ska.Sun import nominal_roll
from sparkles import run_aca_review
from Quaternion import Quat

dec = 89.9
date = '2019:006'
roll = nominal_roll(0, dec, time=date)
att = Quat([0, dec, roll])
print(f'nominal_roll = {roll} == {att.roll}')

obsid = 18000
label = f'obs{obsid}'

kwargs = mod_std_info(att=att, n_guide=8, obsid=obsid, date=date)
aca = get_aca_catalog(**kwargs)

acar = aca.get_review_table()

run_aca_review(acars=[acar], make_html=True, load_name=label,
               report_dir=label, roll_level='all')
```

taldcroft:
And yes, as predicted, I found some oddity in this high-dec case where it finds a roll option if it is labeled as an ER but not when labeled as OR.  Will investigate!

jeanconn:
It seems to me that is possible (when rotating about slightly different
center).

taldcroft:
Actually my code is fine, the issue was that it uses `n_stars` as part of the metric to decide if a new roll is better.  However `n_stars` here is `guide_count` for ORs (roughly the count of 10th mag or brighter stars) and it is `guide_count_9th` for ERs (roughly the count of 9th mag or brighter stars).  

I think these counts are the right ones for evaluating ORs and ERs respectively, but it can be confusing.

So probably just add a test and we're good?

sparkles/roll_optimize.py
```diff
@@ -150,39 +150,46 @@ def get_roll_intervals(self, cand_idxs, roll_nom=None, roll_dev=None,
...
         if roll_nom is None:
-            roll_nom = Ska.Sun.nominal_roll(q_att.ra, q_att.dec, self.date)
+            roll_nom = Ska.Sun.nominal_roll(att.ra, att.dec, self.date)
+            att_nom = Quat([att.ra, att.dec, roll_nom])
+            att_nom_targ = self._calc_targ_from_aca(att_nom, y_off, z_off)
+            roll_nom = att_nom_targ.roll
         if roll_dev is None:
             roll_dev = allowed_rolldev(pitch)
 
         # Ensure roll_nom in range 0 <= roll_nom < 360 to match q_att.roll.
```
jeanconn: Nit: comment references q_att.roll .

sparkles/roll_optimize.py
```diff
@@ -150,39 +150,46 @@ def get_roll_intervals(self, cand_idxs, roll_nom=None, roll_dev=None,
...
         if roll_nom is None or roll_dev is None:
             import Ska.Sun
-            pitch = Ska.Sun.pitch(q_att.ra, q_att.dec, self.date)
+            pitch = Ska.Sun.pitch(att.ra, att.dec, self.date)
         if roll_nom is None:
-            roll_nom = Ska.Sun.nominal_roll(q_att.ra, q_att.dec, self.date)
+            roll_nom = Ska.Sun.nominal_roll(att.ra, att.dec, self.date)
+            att_nom = Quat([att.ra, att.dec, roll_nom])
+            att_nom_targ = self._calc_targ_from_aca(att_nom, y_off, z_off)
+            roll_nom = att_nom_targ.roll
```
jeanconn: Another nit:  docstring should probably mention that roll_nom should be supplied in target frame (not that it is really ever supplied).

jeanconn:
I did another pass through the code and it makes sense to me and I think it is good to go.  Addressing nits optional. 

The logical intervals handling is a little fancy for me, but the rolls look consistently to be supplied in target frame and then transformed to ACA frame to actually check catalogs (which makes sense to me).

taldcroft:
OK, I've update the script and report HTML pages from https://github.com/sot/sparkles/pull/97#issuecomment-468417949.  Please re-review and see if the numbers all seem to make sense.

I excluded a bunch of stars from the initial catalog to make it worse and thus make the roll options more attractive.

sparkles/core.py
```diff
@@ -622,15 +623,20 @@ def get_text_pre(self):
         """
         P2 = -np.log10(self.acqs.calc_p_safe())
         att = self.att
+        att_targ = self.att_targ
         self._base_repr_()  # Hack to set default ``format`` for cols as needed
         catalog = '\n'.join(self.pformat(max_width=-1, max_lines=-1))
         self.acq_count = np.sum(self.acqs['p_acq'])
 
+        att_string = f'ACA RA, Dec, Roll (deg): {att.ra:.6f} {att.dec:.5f} {att.roll:.5f}'
```
jeanconn: Though I suppose I meant to ask if RA really deserved an extra decimal place compared Dec and Roll.

jeanconn:
I am not that surprised that test_get_roll_intervals is now failing.

jeanconn:
I cut the no-longer-true assert statement and pushed to this branch directly.

jeanconn:
I think maybe the roll "mock obsids" in the roll report are still the acar.att.roll values?  A quick search didn't show me where those actually come from so haven't finished with that.

taldcroft:
@jeanconn @jskrist - I've done a bunch more clean-up and I think addressed all issues raised so far.  I made the report page more clear and added the roll options summary table to the roll options review page.  See the sample report pages linked above as they have been updated.

Except didn't do the 6 to 5 decimals places.  I agree that 5 is fine, I just didn't get to it.

PR #102, opened 2019-02-28
Add tests for the geometry checks

jeanconn:
Add tests for the geometry checks


jeanconn:
These are great but they are something.

PR #103, opened 2019-02-28
WIP just an idea for adding SI/Target att to report

jeanconn:
WIP just an idea for adding SI/Target att to report

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -626,11 +626,18 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         catalog = '\n'.join(self.pformat(max_width=-1, max_lines=-1))
         self.acq_count = np.sum(self.acqs['p_acq'])
 
+        aca_att_string = f"RA, Dec, Roll (deg): {att.ra:.6f} {att.dec:.5f} {att.roll:.5f}"
+        if self.is_OR:
+            targ_att_string = (
+                f"(SI/Target Frame {self.att_targ.ra:.3f}, {self.att_targ.dec:.3f}, {self.att_targ.roll:.3f})")
+            att_string = f"{aca_att_string} {targ_att_string}"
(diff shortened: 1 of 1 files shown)


jeanconn:
@jskrist I think this is too busy for the real report, but maybe running with the PR would help to confirm that the numbers you are seeing match my/our understanding that this is just target vs ACA frame stuff?

sparkles/core.py
```diff
@@ -626,11 +626,18 @@ def get_text_pre(self):
         catalog = '\n'.join(self.pformat(max_width=-1, max_lines=-1))
         self.acq_count = np.sum(self.acqs['p_acq'])
 
+        aca_att_string = f"RA, Dec, Roll (deg): {att.ra:.6f} {att.dec:.5f} {att.roll:.5f}"
```
taldcroft:
```
att_labels = [(self.att, 'ACA')] if self.is_ER else [(self.att, 'ACA   '), (self.att_targ, 'Target')]
att_strings = [f'{label} RA, Dec, Roll (deg): {att.ra:.6f} {att.dec:.5f} {att.roll:.5f}' 
               for att, label in att_labels]
att_string = os.linesep().join(att_strings)
```

jskrist:
> @jskrist I think this is too busy for the real report, but maybe running with the PR would help to confirm that the numbers you are seeing match my/our understanding that this is just target vs ACA frame stuff?

With this PR, the values reported in the html as "Roll" are the same as what I am showing for all roll_options (on both ORs and ERs). 👍

jeanconn:
Uh oh, but which "Roll"? 😄

taldcroft:
I just put this idea in #97 ( ed66177).

taldcroft:
The idea here was moved into #97 to make evaluation of results there easier.

PR #107, opened 2019-03-08
Use setuptools_scm for versioning

taldcroft:
See https://github.com/sot/eng_archive/pull/157 for full discussion.

This is a more typical example of what this looks like.


sparkles/__init__.py
```diff
@@ -1,4 +1,19 @@
-__version__ = '4.1.2'
+from pkg_resources import get_distribution
+
+try:
+    _dist = get_distribution(__package__)
+    __version__ = _dist.version
+
+    if not __file__.startswith(_dist.location):
```
taldcroft: Both values `__file__` and `_dist.location` need to be `.lower()` because Windows can return different cases.

setup.py
```diff
@@ -13,7 +11,8 @@
       author='Tom Aldcroft',
       description='Sparkles ACA review package',
       author_email='taldcroft@cfa.harvard.edu',
-      version=__version__,
+      use_scm_version=True,
+      setup_requires=['setuptools_scm', 'setuptools_scm_git_archive'],
```
taldcroft: @javierggt - these two `setup_requires` are sufficient to enable the current build process for Windows:
- Download zip file corresponding to a tagged release
- Unzip
- `python setup.py install` (into an SVN-controlled distribution that is then synced to user laptops)
This will auto-download those from pypi for the build step, so a network is required, but that is not a problem for building.

sparkles/__init__.py
javierggt: If I understand correctly, these are the only lines required to install these packages in the simplest cases. The question is whether we want to set the __version__ variable for use within python anyway.

taldcroft:
This was a proof of concept, closing in favor of #131.

PR #84, opened 2019-02-22
Add caution statement on fewer-than-requested guide stars in OR

jeanconn:
Add caution statement on fewer-than-requested guide stars in OR

PR #117, opened 2019-05-31
Check the number of guides and fids selected vs expected and typical

taldcroft:
This incorporates #84 and expands on the idea, in particular adding a caution that would indicate the likely presence of a MON star in the observation.  It also adds the missing test for #84 plus a few more.

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -856,17 +856,6 @@ Predicted Acq CCD temperature (init) : {self.acqs.t_ccd:.1f}"""
         elif P2 < P2_lim + 1:
             self.add_message('warning', f'P2: {P2:.2f} less than {P2_lim + 1} for {obs_type}')
 
-    def check_n_guide(self):
-        """Check for "typical" number of fids for an OR or ER (3 or 0)
-        """
-        typical_n_guide = 5 if self.is_OR else 8
-        n_guide = len(self.guides)
-        if n_guide != typical_n_guide:
-            msg = f'Catalog has {n_guide} guides but {typical_n_guide} is typical'
-            if self.is_OR and n_guide == 4:
-                msg += f' (could be MON star? Check OR list)'
-            self.add_message('caution', msg)
-
     def check_guide_count(self):
         """
         Check for sufficient guide star fractional count.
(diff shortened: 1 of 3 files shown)


sparkles/tests/test_checks.py
```diff
@@ -56,6 +86,8 @@ def test_guide_count_er():
     assert msg['category'] == 'critical'
     assert 'ER count of 9th' in msg['text']
 
+
+def test_guide_count_er2():
```
taldcroft: Note that as long as the setup is not being repeated it is nicer to split tests into atomic units.

sparkles/tests/test_checks.py
```diff
@@ -40,7 +40,37 @@ def test_check_P2():
     assert 'less than 3.0 for ER' in msg['text']
 
 
-def test_guide_count_er():
+def test_n_guide_check_not_enough_stars():
+    """Test the check that number of guide stars selected is typical"""
```
jeanconn: I think this is checking that the number of guide stars is as-requested.

sparkles/core.py
```diff
@@ -884,6 +884,23 @@ def check_guide_count(self):
...
+            self.add_message(
+                'caution',
+                f'{obs_type} with {n_guide} guides but {self.n_guide} were requested')
+
+        # Caution for any "unusual" guide star request
+        typical_n_guide = 5 if self.is_OR else 8
+        if self.n_guide != typical_n_guide:
+            msg = f'{obs_type} with {n_guide} guides requested but {typical_n_guide} is typical'
+            if self.is_OR and n_guide == 4:
+                msg += f' (likely MON star, check OR list)'
```
jeanconn: Not feedback on the PR, but I note that we don't always have an OR list to check.

sparkles/tests/test_checks.py
taldcroft: Fixed.

sparkles/core.py
```diff
@@ -884,6 +884,23 @@ def check_guide_count(self):
                 'caution',
                 f'{obs_type} with more than {bright_cnt_lim} stars brighter than 6.1.')
 
+        n_guide = len(self.guides)
+
+        # Different number of guide stars than requested
+        if n_guide != self.n_guide:
+            self.add_message(
+                'caution',
```
jeanconn: I'm going back and forth on if this might be worthy of a 'warning'.

sparkles/core.py
taldcroft: Sure, but in practice if you actually see this message on your screen there almost surely be a corresponding OR list somewhere.

jeanconn: Sure, there may be a list, but I *think* that there are no entries in the list for say, pool targets, until after prelim review?  Or maybe I just haven't tried hard enough to find it.   Just an aside.

sparkles/core.py
taldcroft: Could go either way.  I don't think it matters all that much in the end, and perversely a caution will be more visible than warning because they are unusual.  I just want to be done with this and move on to the next thing.  If you want to make a commit to change then that's fine.

PR #118, opened 2019-05-31
Report effective T_ccd when relevant

taldcroft:
I had thought that reporting of effective T_ccd was already in sparkles, but apparently not.


jeanconn:
I like this, but was planning for it to be in ska3-flight and not in the matlab tools currently in testing.

jeanconn:
There is presently debate at today's MPCWG on this issue and if the guideline can be promoted without the updated matlab tools.  The current matlab tools in testing don't have this sparkles fix anyway, so I don't see any reason to hold up the guideline change.

PR #121, opened 2019-06-07
Fix test output directory redux

### LINKED ISSUE #119: test_run_aca_review_function failing from installed package
I'm getting:

sparkles/tests/test_review.py::test_run_aca_review_function FAILED

from my 32bit test environment when run with sparkles.test(verbose=1).  It also complains that it can't write out the error 

PermissionError: [Errno 13] Permission denied: '/export/aca/ska3matlab32/lib/python3.6/site-packages/.cache/v/cache/lastfailed'

back into my environment. 

The test passes if I run it directly in the sparkles repo directory

taldcroft:
Fixes #119 
Closes #120 (supercedes).

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -95,7 +95,7 @@ def run_aca_review(load_name=None, *, acars=None, make_html=True, report_dir=Non
     the list of obsids being processed.  This is handy for development or
     for examining just one obsid.
 
-    If ``outdir`` is not provided then it will be set to ``load_name``.
+    If ``report_dir`` is not provided then it will be set to ``load_name``.
 
     The ``report_level`` arg specifies the message category at which the full
     HTML report for guide and acquisition will be generated for obsids with at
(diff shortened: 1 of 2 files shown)


jeanconn:
Sure.  This works.  I now see the output I expect in
```
pytest-of-jeanconn/pytest-30/test_run_aca_review_function0/
```
with no error run as the user who can't write back into the python libs.  It doesn't do cleanup but I don't think we care.

taldcroft:
> It doesn't do cleanup but I don't think we care.

Normal testing protocol is to let `pytest` do the cleanup.

jeanconn:
Is that just when it gets to certain NUM or some such?  I could go read the docs at some point.  I think we also want to update testr options at some point to keep pytest from doing other outputs in the tree (no:cacheprovider andPYTHONDONTWRITEBYTECODE=1 ?).

taldcroft:
Perfect is the enemy of good!

jeanconn:
I meant at some point, not for this release.   I still don't like any of our testing polluting our configured python environment, and these test actions can throw confusing errors when we don't have write permission back into that directory.

PR #125, opened 2019-08-21
Update for planning limit rename to penalty limit

jeanconn:
Update tests for planning limit rename to penalty limit.


jeanconn:
Needs to go simultaneous to https://github.com/sot/proseco/pull/307

PR #127, opened 2019-09-20
Add GitHub action for flake8 checking

taldcroft:
For this repo the only required ignore is W504 which just means that it *allows* line break after a binary operator but will flag line break before binary operator.  This is the standard convention, but older repos that don't pass can also ignore W503 so that either convention is OK.

I realized that E402 (code intermixed with top-level imports) should always be checked, so I have dropped that from the standard list of ignores.  In cases where it is needed (e.g. `matplotlib.use('Agg')` then put a `# noqa`.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -1,0 +1,19 @@
+name: Python flake8 check
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


taldcroft:
Somehow there was a PEP8 problem that didn't get caught before, but this acts as a nice test that the action is working!  I'm going to merge sans review to keep things rolling since this does not affect the distributed package.

PR #129, opened 2019-10-29
Changes to use versioneer for single-source versioning

taldcroft:
Just trying to see what this would look like.

cc @jzuhone @jeanconn


taldcroft:
We have settled on setuptools_scm for Ska3 packages, so closing this.  

Setuptools_scm seems to be a much smaller footprint change for our 40+ packages that would be updated.  The main downside is adding a developer run-time dependency on setuptools_scm, but that is already in ska3, so not really an issue.  Plan is a new lightweight package `ska_version` to allow this in a package `__init__.py`.
From:
```
__version__ = '1.2.3'
```
to:
```
from ska_version import get_version
__version__ = get_version()
del get_version
```
I note that astropy is going use setuptools_scm, and that this is officially endorsed by the PyPA.

PR #134, opened 2020-01-31
Add clickable call args output and chandra_aca version

taldcroft:
The only problem now is that you need to click on the button twice the first time to get action.  I'll find a javascript expert to figure that out.

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -11,6 +11,7 @@ import traceback
 from pathlib import Path
 import pickle
 from itertools import combinations, chain
+import pprint
 
 import matplotlib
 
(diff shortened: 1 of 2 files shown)


jeanconn:
I was playing with this new feature when looking over the APR0620 test products and I wonder if we need a little be of extra documentation in the args block?   I had forgotten that this is set to force all the stars in the "call args" area with the include_ids and halfws, but those aren't the original "call_args" for the catalogs.  Not sure what makes the most sense to keep from getting confused.

taldcroft:
You won't get confused the second time, and it's pretty obvious on looking at this that not every single catalog has every single guide and acq star forced.

taldcroft:
And if one of us just knocks off #34 one afternoon there will be even less confusion.

jeanconn:
Agreed on #34. 

And regarding confusion, that's probably true.  I figured bringing it up in github this way was probably a good place to track it that it threw me.  For this specific case I thought there could have been something wrong with the APR0620T test products (proseco saving all the includes in the real call args) but quickly confirmed that was not the case.  I also didn't know if you are already using new sparkles in your prelim load reviews (so you wouldn't be surprised by the change if you've been seeing it for weeks).

 I'm not sure about the use case for the call_args + all the includes?  But don't have a strong opinion.

PR #137, opened 2020-02-07
Update regress data for 2020-02 acq model

jeanconn:
Update regress data for 2020-02 acq model


jeanconn:
I can't immediately tell if we've lost any important test coverage.

PR #140, opened 2020-04-10
Add info statement for include / exclude

### LINKED ISSUE #34: It would be useful to mark include/exclude stars with 'info' statements

taldcroft:
## Description

APR2020P seemed like a good opportunity to code this up.

## Testing

- [x] Passes unit tests on MacOS with chandra_aca 4.28.2.dev13+g3c14aa0 installed, fails unrelated tests with current release.
- [x] Functional testing - added new unit tests

Fixes #34

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -904,7 +904,7 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}{t_ccd_eff_acq_msg}""
             self.add_message('warning', f'P2: {P2:.2f} less than {P2_lim + 1} for {obs_type}')
 
     def check_include_exclude(self):
-        """Check for included or excluded guide or acq stars (info)
+        """Check for included or excluded guide or acq stars or fids (info)
 
         """
         call_args = self.call_args
@@ -913,6 +913,9 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}{t_ccd_eff_acq_msg}""
                 ids = call_args.get(f'{action}_ids_{typ}')
                 if ids is not None:
                     msg = f'{action}d {typ} ID(s): {ids}'
+
+                    # Check for halfwidths.  This really only applies to
+                    # include_halfws_acq, but having it here in the loop doesn't hurt.
                     halfws = call_args.get(f'{action}_halfws_{typ}')
                     if halfws is not None:
                         msg = msg + f' halfwidths(s): {halfws}'
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -778,6 +778,7 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}{t_ccd_eff_acq_msg}""
         self.check_acq_p2()
         self.check_guide_count()
         self.check_fid_count()
+        self.check_include_exclude()
 
     def check_guide_geometry(self):
         """Check for guide stars too tightly clustered.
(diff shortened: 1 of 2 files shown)


jeanconn:
As an aside, I don't know one could check out  4.28.2.dev13+g3c14aa0 for reproducibility (seems like 4.29 would work?).  Unrelated to this PR though.

sparkles/core.py
```diff
@@ -902,6 +903,22 @@ def check_acq_p2(self):
         elif P2 < P2_lim + 1:
             self.add_message('warning', f'P2: {P2:.2f} less than {P2_lim + 1} for {obs_type}')
 
+    def check_include_exclude(self):
+        """Check for included or excluded guide or acq stars (info)
+
+        """
+        call_args = self.call_args
+        for typ in ('acq', 'guide', 'fid'):
```
jeanconn, line 911: I suppose since fids aren't supported yet, that the API hasn't been defined.  But this seems reasonable.  Method docstring could mention fids I suppose.

sparkles/core.py
```diff
@@ -902,6 +903,22 @@ def check_acq_p2(self):
...
+            for action in ('include', 'exclude'):
+                ids = call_args.get(f'{action}_ids_{typ}')
+                if ids is not None:
+                    msg = f'{action}d {typ} ID(s): {ids}'
+                    halfws = call_args.get(f'{action}_halfws_{typ}')
```
jeanconn, line 916: This is a little confusing since the only allowed type is include_haflws_acq .  Not sure what is won by having it in the loop.

sparkles/tests/test_checks.py
```diff
@@ -148,6 +149,28 @@ def test_guide_count_er4():
...
+    """Test INFO statement for explicitly included/excluded entries"""
+    stars = StarsTable.empty()
+    stars.add_fake_constellation(n_stars=8, mag=np.linspace(7.0, 8.75, 8))
+    aca = get_aca_catalog(**mod_std_info(obsid=50000, n_fid=0, n_acq=5, n_guide=8),
+                          stars=stars, dark=DARK40,
+                          exclude_ids_acq=100,
+                          include_ids_guide=107,
+                          exclude_ids_guide=[100, 101],
+                          include_ids_acq=[106, 107],
+                          include_halfws_acq=[140, 120],
```
jeanconn, line 162: I've forgotten, for the implicit case where one wants to include an acq but let proseco optimize halfw, does that end up passed from ORViewer as 0 or None?  And what do we want to see for INFO for those case?  Or doesn't matter/don't care?

sparkles/core.py
jeanconn: Ah.  I see, you are winning getting the message in a single line.  That's fine.

taldcroft: The first draft explicitly only checked for `include_halfws_acq`, but the code logic was more complicated so I just chopped it down.

sparkles/tests/test_checks.py
taldcroft: I think we will get a single 0, but I'm not certain (though I'm pretty confident it won't be `None`).  

The upshot is that whatever gets passed (including the argument not being supplied at all), we will be able to unambiguously interpret the output without trouble.  In other words, these new info statements provide a complete record of all include/exclude arguments via `call_args`, and we the reviewers understand the implications.

sparkles/core.py
taldcroft: Yes, when fids are included they will follow the same API.  I wanted to put this in now so we don't need to remember later.  I'll update the docstring though.

PR #149, opened 2021-01-16
Update tests for AGASC supplement disable as needed

taldcroft:
## Description

This disables use of the AGASC supplement mag updates to ensure that regression tests always pass.

## Testing

Using https://github.com/sot/proseco/pull/345 and https://github.com/sot/agasc/pull/56, and the latest `agasc_supplement.h5`

- [x] Passes unit tests on MacOS
- [n/a] Functional testing

### DIFF
--- a/sparkles/tests/test_review.py
+++ b/sparkles/tests/test_review.py
@@ -6,6 +6,7 @@ import pytest
 from proseco import get_aca_catalog
 from proseco.characteristics import aca_t_ccd_penalty_limit
 
+import agasc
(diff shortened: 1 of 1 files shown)


jeanconn (approved):
LGTM, though not entirely clear on the use of the decorator vs context manager.

taldcroft:
It turns out the decorator doesn't play well with the pytest fixture (`tmpdir`). Pytest does some magic to make those fixtures work from what I know.

PR #154, opened 2021-07-02
Fix test regressions with proseco 5.0.0

taldcroft:
## Description

The full catalog merge_cats is more strict / correct than the previous code, so some regressions in sparkles showed up. 

In addition the new bad region ended up changing the regression outputs for the roll options test. This took some time to track down but eventually making the acquisition report and comparing showed the new hot pixels / imposter blocking the 160 arcsec search box in one star.

## Testing

- [x] Passes unit tests on MacOS with proseco 5.0.0
- [x] Passes unit tests on MacOS with https://github.com/sot/proseco/pull/360
- [n/a] Functional testing


sparkles/tests/test_checks.py
```diff
@@ -278,18 +278,18 @@ def test_guide_edge_check():
     col_lim = -(CCD['col_max'] - CCD['col_pad'] - CCD['window_pad'] - dither / 5)
 
     # Set positions just below or above CCD['guide_extra_pad'] in row / col
-    stars.add_fake_star(id=101, mag=8, row=row_lim - 2.9, col=0)
-    stars.add_fake_star(id=102, mag=8, row=row_lim - 3.1, col=0)
-    stars.add_fake_star(id=103, mag=8, row=row_lim - 5.1, col=0)
-    stars.add_fake_star(id=104, mag=8, row=0, col=col_lim + 2.9)
-    stars.add_fake_star(id=105, mag=8, row=0, col=col_lim + 3.1)
-    stars.add_fake_star(id=106, mag=8, row=0, col=col_lim + 5.1)
+    stars.add_fake_star(id=1, mag=8, row=row_lim - 2.9, col=0)
```
taldcroft, line 281: `add_fake_constellation` below was actually adding stars starting from id=106, so there were two id=106 stars. I'm not sure why this was working previously, but the fix is just to not use stars with those ids here.

sparkles/tests/test_review.py
```diff
@@ -60,7 +60,7 @@ def test_review_catalog(tmpdir):
     acar.run_aca_review()
     assert acar.messages == [
         {'text': 'Guide star imposter offset 2.6, limit 2.5 arcsec', 'category': 'warning',
-         'idx': 2},
+         'idx': 4},
```
taldcroft, line 63: Reordering from the full catalog handling.

sparkles/tests/test_review.py
```diff
@@ -309,21 +310,21 @@ def test_roll_options_dec89_9():
     exp[48000] = [' roll   P2  n_stars improvement roll_min roll_max  add_ids   drop_ids',
                   '------ ---- ------- ----------- -------- -------- --------- ---------',
                   '287.25 3.61    0.55        0.00   287.25   287.25        --        --',
-                  '281.00 7.44    6.98        9.64   276.75   285.25 608567744        --',
+                  '281.00 7.24    6.98        9.53   276.75   285.25 608567744        --',
```
taldcroft, line 317: Due to the new bad region.

sparkles/tests/test_review.py
```diff
@@ -305,25 +306,29 @@ def test_roll_options_dec89_9():
...
                   '276.94 3.61    7.54        0.00   276.94   276.94        --        --',
-                  '270.57 7.44    8.00        1.99   266.19   274.94 608567744        --',
                   '277.07 7.25    8.00        1.89   258.19   295.69        --        --',
+                  '270.57 7.16    8.00        1.84   266.19   274.94 608567744        --',
                   '258.19 6.82    8.00        1.68   258.19   262.69 610927224 606601776',
                   '259.69 6.82    8.00        1.68   258.19   262.69 610927224        --']
 
     for obsid in (48000, 18000):
-        kwargs = mod_std_info(att=att, n_guide=8, obsid=obsid, date=date)
+        kwargs = mod_std_info(att=att, n_guide=8, n_fid=0, obsid=obsid, date=date)
```
taldcroft, line 331: Explicitly specify zero fids.

sparkles/tests/test_review.py
jeanconn: When I looked at this I was thinking if we end up adding new regions we could override proseco characteristics differently for regression tests (similar to DARK40) but given where we are in the mission that seemed unlikely to be useful.

taldcroft: That's an option, but it can also be illuminating to see the effects.

sparkles/tests/test_review.py
jeanconn: It wouldn't matter to specify n_fid=3 and n_guide=5 for the OR obsid but this seems simpler if the real point of the two types is to get different roll choices based on the default science target offsets.  For 48000 I wasn't sure if detector should also be specified to be more like every other ER?  But if you change the n_fid default in proseco I guess it doesn't matter.  I suppose I just found it interesting that proseco and sparkles have different strategies about deciding if something is/was an OR.

sparkles/tests/test_checks.py
jeanconn: I thought that was a weakness in `add_fake_constellation`, as it seems like it should just not add ids that are already in the field.  But given it is just a test routine, fine with the strategy to just manually add ids that don't conflict.

sparkles/tests/test_review.py
jeanconn: Yes.  I had planned to just revert characteristics to confirm the thought that it was a second order effect based on your comments that basically 1% of the CCD was effected.

sparkles/tests/test_review.py
javierggt: shouldn't this be an internal constraint?

Just wanted to add some noise 😄

sparkles/tests/test_checks.py
taldcroft: Yes, that's all true.

sparkles/tests/test_review.py
taldcroft: When searching roll options then there is much more likelihood of running into a bad region.

sparkles/tests/test_review.py
jeanconn: By internal constraint, you mean that n_fid + n_guide > 8 should throw an error at the start?  I think Tom's fix in proseco to unset the n_fid default actually does have that effect in most cases.

jeanconn (approved):

sparkles/tests/test_review.py
taldcroft: > I suppose I just found it interesting that proseco and sparkles have different strategies about deciding if something is/was an OR.

Basically proseco does not have the concept of OR/ER at all. There is only one place that pays attention to OR/ER: `get_monitors` which gets monitor window commanding for an existing starcheck catalog text.

PR #156, opened 2021-07-13
Test that roll options work without error on a specific monitor cat

jeanconn:
## Description

Test that get_roll_options runs without error on a specific monitor cat

This is really just a test that duplicates the issue described in https://github.com/sot/proseco/issues/364

### DIFF
--- a/sparkles/tests/test_review.py
+++ b/sparkles/tests/test_review.py
@@ -184,6 +184,24 @@ def test_roll_options_with_include_ids():
     # assert len(acar.roll_options) > 1
 
 
+def test_roll_options_with_monitor_star():
+    """
+    Test case from JamesK using monitor star
+    """
+    kwargs = {'att': [0.32916333, -0.50759709,  0.07481427,  0.79271655],
+              'date': '2021:116:16:42:09.065', 'detector': 'ACIS-I',
+              'dither_acq': (7.9992, 7.9992), 'dither_guide': (7.9992, 7.9992),
+              'man_angle': 86.49934496445843, 'n_acq': 8, 'n_fid': 3, 'n_guide': 5,
+              'obsid': 23050.0, 'sim_offset': 0.0, 'focus_offset': 0.0,
+              't_ccd_acq': -9.067548914167258, 't_ccd_guide': -8.616156814261098,
+              't_ccd_penalty_limit': -6.8,
+              'monitors': [[335.516667, 58.675556, 0., 9.63, 0.]]}
+    aca = get_aca_catalog(**kwargs)
+    acar = aca.get_review_table()
+    acar.run_aca_review()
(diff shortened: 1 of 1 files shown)


sparkles/tests/test_review.py
```diff
@@ -184,6 +184,24 @@ def test_roll_options_with_include_ids():
     # assert len(acar.roll_options) > 1
 
 
+def test_roll_options_with_monitor_star():
+    """
+    Test case from JamesK using monitor star
```
taldcroft, line 189: More like
```
Test that roll optimization succeeds in the case where a monitor star is
specified in the catalog. This tests https://github.com/sot/proseco/issues/365
which fixes https://github.com/sot/proseco/issues/364.
```

sparkles/tests/test_review.py
```diff
@@ -184,6 +184,24 @@ def test_roll_options_with_include_ids():
...
+    """
+    kwargs = {'att': [0.32916333, -0.50759709, 0.07481427, 0.79271655],
+              'date': '2021:116:16:42:09.065', 'detector': 'ACIS-I',
+              'dither_acq': (7.9992, 7.9992), 'dither_guide': (7.9992, 7.9992),
+              'man_angle': 86.49934496445843, 'n_acq': 8, 'n_fid': 3, 'n_guide': 5,
+              'obsid': 23050.0, 'sim_offset': 0.0, 'focus_offset': 0.0,
+              't_ccd_acq': -9.067548914167258, 't_ccd_guide': -8.616156814261098,
+              't_ccd_penalty_limit': -6.8,
+              'monitors': [[335.516667, 58.675556, 0., 9.63, 0.]]}
```
taldcroft, line 198: Use the `MonCoord.RADEC` and `MonFunc.AUTO` enums instead of `0`.

sparkles/tests/test_review.py
```diff
@@ -184,6 +184,24 @@ def test_roll_options_with_include_ids():
...
+    aca = get_aca_catalog(**kwargs)
+    acar = aca.get_review_table()
+    acar.run_aca_review()
+    acar.get_roll_options()
```
taldcroft, line 202: Does the output need to be checked? That is got a catalog with no warnings after `run_aca_review` and that the roll options are reasonable?

jeanconn: That's more a question for you on what you think is due diligence here.  With the previous issue/bug get_roll_options raised an exception so this is an (implicit) test for that (and the test fails as expected on proseco master).  I can certainly add more things to check.

taldcroft: I guess just checking that there are no messages in the `acar` and then adding a comment that getting roll options fails without the fix is fine.

taldcroft (approved):

PR #159, opened 2021-08-18
Fix testing guide count when monitors are present

taldcroft:
## Description

Fix a problem noted on Slack today (search "caution: OR with 5 guides requested but 5 is typical").

## Testing

- [x] Passes unit tests on MacOS, linux (with new unit tests)
- [n/a] Functional testing

Fixes #

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -1038,22 +1038,37 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}{t_ccd_eff_acq_msg}""
                 'caution',
                 f'{obs_type} with more than {bright_cnt_lim} stars brighter than 6.1.')
 
+        # Requested slots for guide stars and mon windows
+        n_guide_or_mon_request = self.call_args['n_guide']
+
+        # Actual guide stars
         n_guide = len(self.guides)
 
-        # Different number of guide stars than requested
-        if n_guide != self.n_guide:
-            self.add_message(
-                'caution',
-                f'{obs_type} with {n_guide} guides but {self.n_guide} were requested')
+        # Actual mon windows. For catalogs from pickles from proseco < 5.0
+        # self.mons might be initialized to a NoneType or not be an attribute so
+        # handle that as 0 monitor windows.
+        try:
+            n_mon = len(self.mons)
+        except (TypeError, AttributeError):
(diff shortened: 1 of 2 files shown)


taldcroft:
@jskrist - this should address the problem you saw today.

sparkles/core.py
```diff
@@ -1038,22 +1038,37 @@ def check_guide_count(self):
...
+        # Actual guide stars
         n_guide = len(self.guides)
 
-        # Different number of guide stars than requested
-        if n_guide != self.n_guide:
-            self.add_message(
-                'caution',
-                f'{obs_type} with {n_guide} guides but {self.n_guide} were requested')
+        # Actual mon windows. For catalogs from pickles from proseco < 5.0
+        # self.mons might be initialized to a NoneType or not be an attribute so
```
jeanconn, line 1048: When loading an pre-proseco 5.0 pickle in an environment with proseco > 5, IIRC I only got a None and not the case of not an attribute (so I assume that the new proseco in the path means that the loaded-pickle objects have the modern mon attributes?).  But the code seems fine.

jeanconn:
Love the new tests too.

jeanconn (approved):

sparkles/core.py
```diff
@@ -1038,22 +1038,37 @@ def check_guide_count(self):
...
-        n_mons = len(self.mons) if self.mons is not None else 0
+        # Different number of guide stars than requested
+        if n_guide + n_mon != n_guide_or_mon_request:
+            if n_mon == 0:
+                # Usual case
+                msg = (f'{obs_type} with {n_guide} guides '
+                       f'but {n_guide_or_mon_request} were requested')
+            else:
+                msg = (f'{obs_type} with {n_guide} guides and {n_mon} monitor(s) '
+                       f'but {n_guide_or_mon_request} guides or mon slots were requested')
```
jskrist, line 1063: This is more verbose, but clearer to me than the tricky bit on line 1068 below, where you are conditionally setting a string to include in the following f-string.  Not sure which is "best-practice" but seeing the two different approaches this close together seems out of sync.

taldcroft: Yeah.. Don't Repeat Yourself is my prime directive. I could do that reasonably in the second case but not here.

PR #161, opened 2021-08-25
Add module to find ER catalog

taldcroft:
## Description

Add module to find ER catalog.

This requires https://github.com/sot/Quaternion/pull/39.

## Testing

- [x] Passes unit tests on MacOS
- [x] Functional testing (new unit tests + [jupyter notebook](https://nbviewer.jupyter.org/urls/cxc.cfa.harvard.edu/mta/ASPECT/proseco/find-er-catalog.ipynb))

Fixes https://github.com/sot/ska-projects/issues/107

### DIFF
--- a/sparkles/tests/test_find_er_catalog.py
+++ b/sparkles/tests/test_find_er_catalog.py
@@ -15,7 +15,7 @@ os.environ[agasc.SUPPLEMENT_ENABLED_ENV] = 'False'
 
 from sparkles.find_er_catalog import (
     get_candidate_stars, find_er_catalog, filter_candidate_stars_on_ccd,
-    get_guide_counts, init_quat_from_attitude)
+    get_guide_counts)
 
 
 # Known tough field: PKS 0023-26 pointing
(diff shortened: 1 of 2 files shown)


taldcroft:
@jeanconn - I'd like to get this in soon so please have a look. It will be better to have it in the release and we can test or develop against it.  We can consider this to be experimental for the moment and it's certainly possible there will be API changes when it gets connected with ORviewer.

jeanconn:
When you say "the release" do you mean the flight one for today?

taldcroft:
No, I just meant the flight release so it isn't just hanging here as a PR for too long. So whenever the next available release.

jeanconn:
Right, the next one is just technically today.  But I would like a little more time to review this.

taldcroft:
I'm using the Polish next, which is the one after the US English next one.

sparkles/tests/test_find_er_catalog.py
```diff
@@ -0,0 +1,145 @@
...
+    get_guide_counts, init_quat_from_attitude)
+
+
+# Known tough field: PKS 0023-26 pointing
+ATT = Quat([0.20668099834, 0.23164729391, 0.002658888173, 0.9505868852])
+DATE = '2021-09-13'
+T_CCD = -8.0
+
+# Get initial catalog at the PKS 0023-26 attitude. Ignore the penalty limit for
+# this work.
```
jeanconn, line 21: Why ignore the penalty limit?

jeanconn: (I suppose I just figure it is unlikely we'll set the penalty limit colder, so it is going to be pretty moot anyway)

sparkles/tests/test_find_er_catalog.py
```diff
@@ -0,0 +1,145 @@
...
+from proseco import get_aca_catalog
+from proseco.tests.test_common import mod_std_info
+import numpy as np
+from Quaternion import Quat
+import Ska.Sun
+
+from sparkles.find_er_catalog import (
+    get_candidate_stars, find_er_catalog, filter_candidate_stars_on_ccd,
+    get_guide_counts, init_quat_from_attitude)
+
```
jeanconn, line 13: I assume this also needs something like the `os.environ[agasc.SUPPLEMENT_ENABLED_ENV] = 'False'` that is used in test_review.py .

PR #162, opened 2021-09-09
Update bright limit check to use 5.2 mag

jeanconn:
## Description

Update bright limit check to use 5.2 mag

## Testing

- [x] Passes unit tests on linux
- [x] Functional testing - updated unit test is a functional test


taldcroft (approved):
LGTM.

jeanconn:
I forgot this test won't pass until https://github.com/sot/proseco/pull/368 goes in to update the star selection.

PR #166, opened 2021-12-20
Add a check for two tracked things that are within 12 pixels.

jeanconn:
## Description

Add a check for two tracked things that are within 12 pixels.  The PEA will drop readout windows that overlap in pixel space, so tracked types should not overlap in pixel space.  This PR adds a check that the center points of any tracked objects are more than 12 pixels apart.

This was noticed in obsids 45890 and 45884 in NOV2921A.

## Testing

- [x] Passes unit tests on linux
- [x] Functional testing - adds unit test

### DIFF
--- a/sparkles/tests/test_checks.py
+++ b/sparkles/tests/test_checks.py
@@ -347,21 +347,34 @@ def test_guide_edge_check():
     acar = ACAReviewTable(aca)
     acar.check_catalog()
 
+    # For this constructed test we're interested in the edge marging warnings
+    # and don't care about the warnings about overlap.
     assert acar.messages == [
-        {'category': 'critical',
-         'idx': 5,
-         'text': 'Less than 3.0 pix edge margin row lim 495.4 val 492.5 delta 2.9'},
-        {'category': 'info',
-         'idx': 6,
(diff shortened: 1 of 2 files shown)


sparkles/tests/test_checks.py
```diff
@@ -331,21 +347,34 @@ def test_guide_edge_check():
     acar = ACAReviewTable(aca)
     acar.check_catalog()
 
+    # For this constructed test we're interested in the edge marging warnings
```
taldcroft, line 350: I would probably just fix the star positions so they don't overlap. I.e. `col=0` then `col=100`, `col=200` and same for `row` later.

taldcroft (changes requested):
Looks basically good but one suggested change in the unrelated test.

sparkles/tests/test_checks.py
jeanconn: OK.  I think I started with that and then figured that if I changed the test more fundamentally, I'd need to justify any changes (and that it was still really doing the same thing) more than just acknowledging the new warnings?  But we can go that way.  Also a typo in my comment apparently.

taldcroft: I'd say it is making the test better since it introduces variations in a parameter that should be independent.

taldcroft (approved):

PR #169, opened 2022-03-24
Add mag for candidates + fix minor plot() bug

taldcroft:
## Description

This adds text that indicates the magnitude of candidate stars in a sparkles plot. This was useful in the Jiang DDT evaluations.

It also fixes a bug I noticed in the `plot` method where the input `ax` arg was being ignored. The jupyter notebook run was a test of this.

## Interface impact

This changes the plot output to add more information but otherwise does not change any impact of code execution.

## Testing

- [x] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing

### Functional testing

```
import matplotlib.pyplot as plt
%matplotlib 
# %matplotlib widget in jupyter notebook
aca = get_aca_catalog(8008)
acar = aca.get_review_table()
fig, ax = plt.subplots(figsize=(6, 6))
acar.plot(ax=ax)  # did this a second time in another cell and it plotted in the original ax
```

image

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -730,7 +730,8 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         :param kwargs: other keyword args for plot_stars
         """
         fig = super().plot(ax, **kwargs)
-        ax = fig.gca()
+        if ax is None:
+            ax = fig.gca()
 
         # Increase plot bounds to allow seeing stars within a 750 pixel radius
         ax.set_xlim(-770, 770)  # pixels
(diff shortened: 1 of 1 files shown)


taldcroft:
@jeanconn @javierggt - let's aim to get this into 2022.3.

jeanconn:
This seems a little helpful, but for me, the bigger issue was just not being able to tell if the star was useful for guide (because I thought the current highlight was all potential guide or acq candidates outside the CCD FOV).  Maybe different color coding?

sparkles/core.py
```diff
@@ -730,7 +730,8 @@ def plot(self, ax=None, **kwargs):
         :param kwargs: other keyword args for plot_stars
         """
         fig = super().plot(ax, **kwargs)
-        ax = fig.gca()
+        if ax is None:
```
jeanconn, line 733: Good catch.

jeanconn (approved):

PR #173, opened 2022-06-01
Tweaks to sparkles / yoshi based on cycle 24 work

taldcroft:
## Description

This PR has a few small updates which facilitate the feasibility studies of cycle 24 proposed targets.

- Add target name to the report output
- Fix roll dev logic so `max_roll_dev` is really the max roll deviation when supplied.
  -  It turns out there was a mistake in the API and one method (used only internally) had a `roll_dev` parameter that was  basically redundant.
  - I made this deprecated and added a FutureWarning if it is supplied.
- Fixed some doc strings mistakes
- Allowed supplying additional proseco keyword args to `run_one_yoshi` and `convert_yoshi_to_proseco_params`. For the cycle 24 studies this allows passing `target_name` directly.

## Interface impacts

`roll_dev` arg to `AcaReviewTable.get_roll_intervals()` method is now deprecated and will raise a `FutureWarning` if supplied.

### Unit tests

- [x] Mac

- [x] Linux:

### Functional tests

Works with https://gist.github.com/taldcroft/92d6d92ed80316a9a66dffbf9379eb52.

### DIFF
--- a/sparkles/tests/test_yoshi.py
+++ b/sparkles/tests/test_yoshi.py
@@ -83,12 +83,12 @@ def test_get_params():
         "ra_targ": 149.91616666666664,
         "roll_targ": 62.01050867568485,
         "sim_offset": 0,
+        "target_name": "C-COSMOS",
     }
     assert_dict_equal(params, exp)
 
     params_proseco = convert_yoshi_to_proseco_params(
-        **params, obsid=8008, t_ccd=-10, man_angle=5.0, target_name="Target name",
-
+        **params, obsid=8008, t_ccd=-10, man_angle=5.0,
     )
     exp_proseco = {
         "att": Quat([0.15017923, 0.49292814, 0.83025727, 0.21246392]),
@@ -103,7 +103,7 @@ def test_get_params():
         "obsid": 8008,
         "sim_offset": 0,
         "t_ccd": -10,
-        "target_name": "Target name",
+        "target_name": "C-COSMOS",
     }
     assert_dict_equal(params_proseco, exp_proseco)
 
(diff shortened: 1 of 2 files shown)


sparkles/roll_optimize.py
```diff
@@ -120,10 +120,18 @@ def get_roll_intervals(self, cand_idxs, d_roll=None, roll_dev=None,
...
-        :param z_off: Z offset (deg, sign per OR-list convention)
         :param d_roll: step size for examining roll range (deg, default=0.25 for
                        uniq_ids and 0.5 for uniform)
+        :param roll_dev: roll deviation about nominal to sample (DEPRECATED,
+                use max_roll_dev instead)
+        :param method: method for determining roll intervals ('uniq_ids' | 'uniform')
+            The 'uniq_ids' method is a faster method that frequently finds an
+            acceptable roll option, while 'uniform' is a brute-force search of
+            the entire roll range at ``d_roll`` increments. If not provided, the
+            default is to try *both* methods in order, stopping when an
```
jeanconn, line 131: Should the default then be None in the definition?

jeanconn:
Should it also just get the target name if available in get_yoshi_params_from_ocat?

taldcroft:
> Should it also just get the target name if available in get_yoshi_params_from_ocat?

Yes, done.

taldcroft:
@jeanconn - I fixed the issues you noted.

jeanconn:
Thanks.  I also ran this on the somewhat stressing case of obsid 17892 which has no target name.  Did not cause any errors or issues with output.  I note that in general that convert_yoshi_to_proseco_params doesn't behave well for old observations (< 2012) because chandra_aca.drift complains that the dynamic stuff doesn't work for historical observations that old. But that's not an issue for this code and not an issue for the way we are likely to use sparkles.

PR #174, opened 2022-06-08
Require minimum number of anchor stars for dyn bgd bonus

taldcroft:
## Description

When adding a temperature delta for the dynamic background bonus, require that at least 3 stars are so-called "anchor stars" with no bonus applied. These anchor stars are evaluated as for the legacy background with no bonus.

## Interface impacts

None

### Unit tests

- [x] Mac

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -38,6 +38,13 @@ from .roll_optimize import RollOptimizeMixin
 CACHE = {}
 FILEDIR = Path(__file__).parent
 
+# Minimum number of "anchor stars" that are always evaluated *without* the bonus
+# from dynamic background when dyn_bgd_n_faint > 0. This is mostly to avoid the
+# situation where 4 stars are selected and 2 are faint bonus stars. In this case
+# there would be only 2 anchor stars that ensure good tracking even without
+# dyn bgd.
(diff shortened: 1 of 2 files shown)


sparkles/core.py
```diff
@@ -38,6 +38,13 @@
 CACHE = {}
 FILEDIR = Path(__file__).parent
 
+# Minimum number of "anchor stars" that are always evaluated *without* the bonus
+# from dynamic background when dyn_bgd_n_faint > 0. This is mostly to avoid the
+# situation where 4 stars are selected and 2 are faint bonus stars. In this case
+# there would be only 2 anchor stars that ensure good tracking even without
+# dyn bgd.
+MIN_DYN_BGD_ANCHOR_STARS = 3
```
jeanconn, line 46: This makes sense, but somewhat defeats having dyn_bgd_n_faint be option-driven.  But if you go with n_guide - dyn_bgd_n_faint there are separate struggles with the corner cases.  Maybe it could be n_guide - dyn_bgd_n_faint with a critical if n_guide != 5 and dyn_bgd_n_faint > 0.

PR #179, opened 2022-10-17
Update sparkles.yoshi to use get_target_aimpoint

jeanconn:
## Description
Update sparkles.yoshi to use get_target_aimpoint . get_target_aimpoint uses the zero offset aimpoint (ZOA) table to get the ZOA locations.  This change makes the sparkles.yoshi code robust to updates to the ZOA table now and for future cycles.

## Interface impacts

Legacy star catalogs may change.

### Unit tests

- [x] Mac

- [x] Mac

### Functional tests

No functional tests.


sparkles/yoshi.py
```diff
@@ -53,15 +53,15 @@ def get_yoshi_params_from_ocat(obsid, obs_date=None, web_ocat=True):
...
-    aimpoints = {
-        "ACIS-I": (970.0, 975.0, 3),
-        "ACIS-S": (210.0, 520.0, 7),
-        "HRC-S": (2195.0, 8915.0, 2),
-        "HRC-I": (7590.0, 7745.0, 0),
-    }
-    chip_x, chip_y, chip_id = aimpoints[ocat["instr"]]
+    # The zero offset aimpoint table starts at cycle 15.
+    # For ancient obsids, just use the earliest in the table.
+    cycle = np.max([15, int(ocat['prop_cycle'])])
```
jeanconn, line 58: I wasn't exactly sure what to do for obsids that pre-date the zero-offset aimpoint table process, so I just set them to have cycle=15 for now.  I don't see this as a particularly useful case for to worry about for this code.

taldcroft: Good enough!

taldcroft (approved):
The code looks good. The only thing you need to do here (and in every PR) is provide a useful description that will make sense to a reader in 3 years. What is the patch doing (meaningful code change summary) and WHY.

sparkles/yoshi.py
```diff
@@ -53,15 +53,15 @@ def get_yoshi_params_from_ocat(obsid, obs_date=None, web_ocat=True):
...
+    chip_x, chip_y, chip_id = get_target_aimpoint(
+        date=obs_date,
+        cycle=cycle,
+        detector=ocat['instr'],
+        too=ocat['type'] in ['TOO', 'DDT'])
```
jeanconn, line 63: @taldcroft should I actually set this to just DDT? It is a bit in the weeds but I thought for the current aimpoint change we were leaving the TOOs with their proposal cycle aimpoints.  That's not consistent with how chandra_aca.drift.get_target_aimpoint handles them if too=True (    :param too: boolean. If target is TOO use current cycle not proposal cycle.).

taldcroft: I have no idea and I am not sure we can codify the upstream rules in a useful way. So I'd suggest just adding `cycle=None` as a function parameter and then do:
```
if cycle is None:
    cycle = np.max([15, int(ocat['prop_cycle'])])
```
And with that do not set the `too` kwarg at all in the `get_target_aimpoint` call. So this is bring explicit control of the cycle up to this level in the very rare circumstance that we actually care.

jeanconn: Do you mean to have "cycle" in that fashion mean more "requested aimpoint cycle"?  I don't think that helps as then it is still on ACA to figure out what it should be.  I think instead we do want to codify this and just do whatever SOT MP is doing.

taldcroft: Mostly I don't quite understand how this would really matter for ACA. If it does then it seems we just ask SOT MP what is the correct cycle to use.

jeanconn: I agree but I would prefer that the most default answer is also really close to being right.  With regard to implementation my only concern with using "cycle=" as an override was that I usually forget which one we're in!

taldcroft: OK well you got a long answer from Josh with a ton of information so hopefully that is enough for the code.

taldcroft: In other words whatever you decide is fine.

jeanconn: I went with your suggestion.

taldcroft (approved):

sparkles/yoshi.py
jeanconn: But *gah* I forgot to remove the too kwarg setting.

jeanconn: I've added one more commit to do that and will rerun the unit tests.

jeanconn:
Added the commit to not pass the TOO kwarg, redid run of unit tests, and re-requested review (though I think it should be trivial at this point).

taldcroft (approved):

PR #182, opened 2023-03-06
Use creep-away to downgrade guide count and cluster critical warns

jeanconn:
## Description

For observations with man_angle_next <= 5deg:

- downgrade cluster warnings from critical to warning
- downgrade guide count warning in 3.5 to 4.0 range from critical to warning

## Interface impacts

Sparkles output updated to include man_angle_next .

## Testing

Tested with ska3

### Unit tests

- [x] Mac

Independent check of unit tests by @taldcroft 
- [x] Mac 

### Functional tests

Unit tests updated to include new guide_count and cluster cases with man_angle_next == 5.

PR #183, opened 2023-03-28
Check for 4x4 dither for 3.5 to 4.0 guide_count

jeanconn:
## Description

Add a requirement / check that fields with low guide count (3.5 <= guide_count < 4) also have reduced (4x4) dither unless we're evaluating with dynamic background.  This implementation infers if we are using dynamic background from dyn_bgd_n_faint.  I don't know if that is the best thing or if we should have something separate to control this like an env VAR.

Fixes #

### Unit tests

- [x] Mac

- [x] Mac

### Functional tests

Unit tests updated to have some that apply to this new check.

### DIFF
--- a/sparkles/tests/test_checks.py
+++ b/sparkles/tests/test_checks.py
@@ -344,6 +344,52 @@ def test_low_guide_count_creep_away():
         {'text': 'OR with 4 guides but 5 were requested', 'category': 'caution'}]
 
 
+def test_reduced_dither_low_guide_count():
+
+    # Set a scenario with guide_count in the 3.5 to 4.0 range
+    # dither == 4x4 to get no warnings
+    stars = StarsTable.empty()
+    stars.add_fake_constellation(n_stars=5, mag=[7.0, 7.0, 7.0, 10.2, 10.3])
+    aca = get_aca_catalog(**mod_std_info(n_fid=3, n_guide=5, obsid=1,
+                                         dyn_bgd_n_faint=0, dither=(4, 4)),
+                          stars=stars, dark=DARK40,
+                          raise_exc=True)
+    acar = ACAReviewTable(aca)
+    acar.check_dither()
+    assert len(acar.messages) == 0
+
+
+def test_not_reduced_dither_low_guide_count():
+    # Set a scenario with guide_count in the 3.5 to 4.0 range and
+    # implicit dyn_bgd_n_faint=0 and dither > 4x4 to get a critical warning
+    stars = StarsTable.empty()
+    stars.add_fake_constellation(n_stars=5, mag=[7.0, 7.0, 7.0, 10.2, 10.3])
+    aca = get_aca_catalog(**mod_std_info(n_fid=3, n_guide=5, obsid=1,
+                                         dyn_bgd_n_faint=0, dither=(8, 8)),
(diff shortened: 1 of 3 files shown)


sparkles/tests/test_checks.py
```diff
@@ -321,19 +321,47 @@ def test_guide_count_or():
...
     stars.add_fake_constellation(n_stars=5, mag=[7.0, 7.0, 7.0, 10.2, 10.3])
-    aca = get_aca_catalog(**mod_std_info(n_fid=3, n_guide=5, obsid=1, man_angle_next=5),
+    aca = get_aca_catalog(**mod_std_info(n_fid=3, n_guide=5, obsid=1,
+                                         dyn_bgd_n_faint=0, dither=(4, 4)),
                           stars=stars, dark=DARK40,
                           raise_exc=True)
     aca = ACAReviewTable(aca)
-    aca.check_guide_count()
+    aca.check_dither()
+    assert len(aca.messages) == 0
```
taldcroft, line 337: Won't there be a warning if guide_count < 4.0 with #182 merged? Maybe this PR should be based off that branch?

taldcroft: In any case it would be good to add an assertion that the guide count is really in the expected range since the fractional guide count definition could change in the future.

sparkles/tests/test_checks.py
```diff
@@ -321,19 +321,47 @@ def test_guide_count_or():
         {'text': 'OR count of guide stars 3.65 < 4.0', 'category': 'critical'},
         {'text': 'OR with 4 guides but 5 were requested', 'category': 'caution'}]
 
-    # Set a scenario with guide_count in the 3.5 to 4.0 range and *but* with
-    # a creep away (maneuver angle <= 5), and confirm that is just a warning
-    # (not critical).
+
+def test_dither_low_guide_count():
```
taldcroft, line 325: Make each of these individual tests into a distinct test function with appropriate docstring.

taldcroft (changes requested):
As mentioned inline, perhaps this needs to be based off of #182 since that could impact messages that appear in the low guide_count regime.

sparkles/tests/test_yoshi.py
```diff
@@ -46,7 +46,7 @@ def test_run_one_yoshi():
         "ra_aca": 238.96459762180638,
         "dec_aca": 66.400811774068146,
         "roll_aca": 197.20855489084187,
-        "n_critical": 2,
+        "n_critical": 3,
```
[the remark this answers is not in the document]
jeanconn: 20562 uses as in the yoshi run test as configured has guide count 3.86, so it gets a new critical warning that the dither is 8x8 (> 4x4).  I didn't see a useful place to capture that in the code comments; I think it should be understood the regress outputs may need to just be updated when warnings are changed or guide_count calculation is modified.

jeanconn:
I've updated this per comments.  Thanks!

sparkles/tests/test_yoshi.py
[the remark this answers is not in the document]
taldcroft: OK makes sense.

taldcroft (approved):
Look good now!

PR #185, opened 2023-04-18
Handle updated pitch roll table in roll optimizations

jeanconn:
## Description

Update to support the pitch/roll table in chandra_models 3.48 via ska_sun.

## Testing
Tested with ska_sun master at [e43f1e1[e43f1e1](https://github.com/sot/ska_sun/commit/e43f1e13b082e78d43b7e1f2f23164c00fa44eaf)

### Unit tests

- [x] Mac

- [x] Mac (using ska3-dev env with some other packages at current master)

### Functional tests

TA got some reasonable results with this script which gets a catalog for an attitude around pitch=40.
```
from ska_sun import apply_sun_pitch_yaw, pitch, nominal_roll
import sparkles
from proseco.tests.test_common import mod_std_info
from Quaternion import Quat

print(sparkles.__version__)

ra = dec = 0.0
date = '2023:001'
roll = nominal_roll(ra, dec, date)
pitch0 = pitch(ra, dec, date)
print(pitch0)

att0 = Quat([ra, dec, roll])
att = apply_sun_pitch_yaw(att0, pitch=40)
print(pitch(att.ra, att.dec, date))

aca = get_aca_catalog(**mod_std_info(att=att, date=date))
acar = aca.get_review_table()

acar.loud = True
acar.get_roll_options()

acar.run_aca_review(make_html=True, roll_level="all", report_dir="pr-185-play")
```


sparkles/roll_optimize.py
```diff
@@ -185,7 +185,10 @@ def get_ids_list(roll_offsets):
                 FutureWarning
             )
         else:
-            roll_dev = Ska.Sun.allowed_rolldev(pitch) if max_roll_dev is None else max_roll_dev
+            # allowed_rolldev returns -1.0 for not-allowed pitch.
+            # Just clip to 0 in that case in the context of this function.
+            roll_dev = (ska_sun.allowed_rolldev(pitch).clip(0) if max_roll_dev is None
```
jeanconn, line 190: The new behavior where allowed_rolldev returns -1.0 doesn't work here directly without modifying the return value, which I did by just clipping.  But I'm not sure if if this block needs rewrite, the whole function needs rewrite, or if the -1.0 return from allowed_rolldev should be reconsidered.

taldcroft: This fix does not actually run for a pitch which is outside the allowed pitch range, failing in the same way that current flight code does in this situation. I took a different approach that seems to work and give expected results.

jeanconn: Great! I like your logic better than my hack. Though my fix worked for me for the test_roll_outside_range test that was failing, so I'm a bit confused about it not actually working for you.  Hmm.  No matter.

taldcroft:
LGTM.
!

taldcroft (approved):
LGTM!

PR #186, opened 2023-05-25
Fix pickling

taldcroft:
## Description

It was noted by @jskrist that the `guide_count` attribute of an `ACAReviewTable` object was getting lost when using this object with parallel processing. The root cause was the same lost attribute when the object was pickled and unpickled.

This PR fixes that particular problem along with a couple of other similar problems that would have been an issue.

Also:
- Adds tests for pickling and copying that were copied almost verbatim from `proseco`. 
- `__init__()` function was slightly re-organized for clarity.
- `is_OR()` property was updated to use a more standard idiom so that the private `self._is_OR` attribute does not require explicit initialization.

Note:
- Sparkles `master` (and this branch) currently require dev versions of packages, at least `chandra_aca` and `cheta` to pass tests.

## Interface impacts

None

### Unit tests

- [ ] No unit tests
- [x] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

From @jskrist : 

I also checked out the branch on my Windows machine, and manually ran the checks in the test_pickle() function (on an independent ACACatalog), as the tests don't run correctly in my Windows environment. Additionally, I ran the code I wrote to parallelize the run_aca_review() step with the new changes and can confirm that with this fix, the multiprocessing code is now returning expected results instead of erroring.

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -528,27 +528,29 @@ class ACAReviewTable(ACATable, RollOptimizeMixin):
         if len(self.colnames) == 0:
             return
 
+        # Init roll option attrs. Note that an instance might be initialized from
+        # another acar instance that already has these attrs set, so we need to
+        # explicitly set them here.
         self.is_roll_option = is_roll_option
+        self.roll_options = None
+        self.roll_info = None
 
         # Add row and col columns from yag/zag, if not already there.
         self.add_row_col()
 
-        self.context = {}  # Jinja2 context for output HTML review
         self.messages = MessagesList()  # Warning messages
+
+        # Instance attributes that won't survive pickling
+        self.context = {}  # Jinja2 context for output HTML review
         self.loud = loud
-        self.roll_options = None
-        self.roll_info = None
         self.preview_dir = None
         self.obsid_dir = None
         self.roll_options_table = None
-        self.acq_count = None
-        self._is_OR = None
 
         # Input obsid could be a string repr of a number that might have have
         # up to 2 decimal points.  This is the case when obsid is taken from the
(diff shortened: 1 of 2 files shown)


jskrist (approved):
The code changes look reasonable to me.  I also checked out the branch on my Windows machine, and manually ran the checks in the `test_pickle()` function (on an independent ACACatalog), as the tests don't run correctly in my Windows environment.  Additionally, I ran the code I wrote to parallelize the `run_aca_review()` step with the new changes and can confirm that with this fix, the multiprocessing code is now returning expected results instead of erroring.

The only minor question I had is related to using `pickle.dumps`/`pickle.loads` in both new tests as opposed to the binary versions of those functions.  I'm just not sure which method multiprocessing might be using under the hood, or if there are any differences between the two versions that would warrant the use of one over the other.

jskrist:
> @jskrist - I checked the multiprocessing docs and by default is uses `pickle` with the default protocol, which is currently 4 (see https://docs.python.org/3/library/pickle.html#data-stream-format). Is that what you mean by the binary version?

Thanks for looking into this.  My question was even simpler than that.  essentially, is there any meaningful difference between `pickle.dumps` and `pickle.dump` for this test?  Do we have confidence that using `loads(dumps())` is equivalent to using `load(dump())`?

taldcroft:
@jskrist - yes, `dumps()` is equivalent to `dump()` : https://docs.python.org/3/library/marshal.html#marshal.dumps : "Return the bytes object that would be written to a file by dump(value, file)"

taldcroft:
@jskrist - and thanks for the functional testing, I put that into the top description.

PR #191, opened 2023-06-16
WIP: Whitespace fix for black

jeanconn:
## Description

Fixes #

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

Independent check of unit tests by [REVIEWER NAME]
- [ ] [PLATFORM]:

### Functional tests

No functional testing.

PR #192, opened 2023-06-22
Fix issue computing t_ccds_bonus and refactor into function

taldcroft:
## Description

The previous implementation had a bug where the `t_ccds_bonus` property was ordered with respect to `sorted(mags)` instead of `mags`. In most cases the guide star mags are roughly sorted in ascending order, but this is not always the case.

An unrelated update was applied to the pre-commit config to get that working.

This also includes the formatting fix in #191.

## Interface impacts

Adds a new function `get_t_ccds_bonus` to the top-level public interface. This can be used in starcheck.

## Testing

Added new unit tests that give better coverage of this functionality.

### Unit tests

- [ ] No unit tests
- [x] Mac
- [ ] Linux
- [ ] Windows

- [x] Mac

### Functional tests

Unit tests added as functional tests.

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -46,6 +46,34 @@ MIN_DYN_BGD_ANCHOR_STARS = 3
 CREEP_AWAY_THRESHOLD = 5.0
 
 
+def get_t_ccds_bonus(mags, t_ccd, dyn_bgd_n_faint, dyn_bgd_dt_ccd):
+    """Return array of t_ccds with dynamic background bonus applied.
+
+    This adds ``dyn_bgd_dt_ccd`` to the effective CCD temperature for the
+    ``dyn_bgd_n_faint`` faintest stars, ensuring that at least MIN_DYN_BGD_ANCHOR_STARS
+    are evaluated without the bonus. See:
+    https://nbviewer.org/urls/cxc.harvard.edu/mta/ASPECT/ipynb/misc/guide-count-dyn-bgd.ipynb
+
+    :param mags: array of star magnitudes
+    :param t_ccd: single T_ccd value (degC)
+    :param dyn_bgd_n_faint: number of faintest stars to apply bonus
+    :returns: array of t_ccds (matching ``mags``) with dynamic background bonus applied
+    """
+    t_ccds = np.full_like(mags, t_ccd)
+
+    # If no bonus stars then just return the input t_ccd broadcast to all stars
+    if dyn_bgd_n_faint == 0:
+        return t_ccds
+
+    idxs = np.argsort(mags)
+    n_faint = min(dyn_bgd_n_faint, len(t_ccds))
+    idx_bonus = max(len(t_ccds) - n_faint, MIN_DYN_BGD_ANCHOR_STARS)
+    for idx in idxs[idx_bonus:]:
+        t_ccds[idx] += dyn_bgd_dt_ccd
+
+    return t_ccds
+
+
 def main(sys_args=None):
     """Command line interface to preview_load()"""
     import argparse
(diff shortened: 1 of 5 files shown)


sparkles/core.py
```diff
@@ -46,6 +46,34 @@
...
+
+    # If no bonus stars then just return the input t_ccd broadcast to all stars
+    if dyn_bgd_n_faint == 0:
+        return t_ccds
+
+    idxs = np.argsort(mags)
+    n_faint = min(dyn_bgd_n_faint, len(t_ccds))
+    idx_bonus = max(len(t_ccds) - n_faint, MIN_DYN_BGD_ANCHOR_STARS)
+    for idx in idxs[idx_bonus:]:
+        t_ccds[idx] += dyn_bgd_dt_ccd
```
jskrist, line 72: Why did this logic switch to using a for loop?  does the previous syntax not work in this case (I don't understand numpy well enough to know)?

`t_ccds[idxs[idx_bonus:]] += dyn_bgd_dt_ccd`

jeanconn: I find the loop more readable once we get to needing to nest the index.

sparkles/core.py
```diff
@@ -700,19 +730,12 @@ def __init__(self, *args, **kwargs):
...
-                # Apply the dynamic background t_ccd bonus to the
-                # dyn_bgd_n_faint faintest stars, ensuring that at least
-                # MIN_DYN_BGD_ANCHOR_STARS are evaluated without the bonus. See:
-                # https://nbviewer.org/urls/cxc.harvard.edu/mta/ASPECT/ipynb/misc/guide-count-dyn-bgd.ipynb
-                mags = np.sort(mags)
-                n_faint = min(self.dyn_bgd_n_faint, len(t_ccds))
-                idx_bonus = max(len(t_ccds) - n_faint, MIN_DYN_BGD_ANCHOR_STARS)
-                t_ccds[idx_bonus:] += self.dyn_bgd_dt_ccd
-            self._t_ccds_bonus = t_ccds
+            self._t_ccds_bonus = get_t_ccds_bonus(
```
jeanconn, line 733: My only concern is that self.mags is not guaranteed to be in the same order as self.guides['mag'].  This isn't a logical issue but suggests a bit to me that the t_ccds_bonus should be an attribute of self.guides not self.

jeanconn: Given urgency and that having the bonus "top level" is not a change in this PR, this is just an aside for future consideration.

jeanconn (approved):

sparkles/core.py
taldcroft: My first iteration was exactly like the @jskrist version, but I also thought it was less readable than a loop. I also noted that corner cases like where `idx_bonus` is beyond the end of the array might need testing when done as fancy indexing instead of the plain loop. (I tested, it works...).

sparkles/core.py
jeanconn: Regarding "That's not a thing in this code as far as I can see." I agree.  But note that there is a "self.mags" that is the mags of the full aca review table that is distinct from self.guides['mags'].

taldcroft: You probably mean `self["mag"]`?

jeanconn: By "having the bonus 'top level'" I just meant that the t_ccds_bonus property is on self and not self.guides and that is not a change introduced in this PR.  I think we're on the same page here.

jeanconn: And "You probably mean self["mag"]?" and yes, I was being imprecise.  I did mean self['mag'] not self.mags.

PR #197, opened 2023-09-28
Update tests for AGASC >= 1.8 and lazy import matplotlib

taldcroft:
## Description

This updates the unit tests to use AGASC >= 1.8 (including release candidates) by default. A number of tests are regression tests that require version 1.7 to obtain the exact same outputs, so those are fixed at 1.7.

In addition a change was made to import matplotlib only when needed for plotting. Sparkles is commonly used only to evaluate the catalog without making a report, so in that case this change saves around 20 Mb of memory and improves performance somewhat.

### Requires

- https://github.com/sot/agasc/pull/155
- https://github.com/sot/proseco/pull/387

## Interface impacts

None

### Unit tests

Using a ska3 dev environment with the required development packages installed and `proseco_agasc_1p8rc4.h5` in `$SKA/data/agasc`:
- [x] Mac

- [x] Linux

### Functional tests

Ran `sparkles https://occweb...MAY1523A_proseco.pkl.gz --open-html` and got the expected result.

### DIFF
--- a/sparkles/tests/test_yoshi.py
+++ b/sparkles/tests/test_yoshi.py
@@ -1,4 +1,3 @@
-import agasc
 import numpy as np
 import pytest
 from mica.archive.tests.test_cda import HAS_WEB_SERVICES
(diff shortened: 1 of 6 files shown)


sparkles/tests/test_yoshi.py
```diff
@@ -148,7 +148,7 @@ def test_get_params_use_cycle():
 
 
 @pytest.mark.skipif(not HAS_WEB_SERVICES, reason="No web services available")
-def test_acar_from_ocat(monkeypatch):
+def test_acar_from_ocat(proseco_agasc_1p7, monkeypatch):
```
jeanconn, line 151: I note that this test probably doesn't need to use monkeypatch and explicitly disable the supplement (that should already be happening via autouse)-- but that's not an issue with this PR.

jeanconn:
This looks good to me, but I'm hesitant to approve it until the proseco and agasc PRs are finalized.  As an aside I note the unit testing section notes that this passes tests with proseco_agasc_1p8rc4.h5 . I also note explicitly that the 1p8 test _won't_ pass unless there is a 1p8 file .

taldcroft:
>  I also note explicitly that the 1p8 test won't pass unless there is a 1p8 file .

Yes, that was intended to ensure that testing environments are configured to cover testing 1p8 (sooner rather than later).

jeanconn (approved):

PR #199, opened 2023-10-18
Add too-cold check for fid lights at acquisition

jeanconn:
## Description

Add a caution warning for acquisition temperature between -13.5 and -14, and a critical for acquisition temperature < -14C.

### Unit tests

- [X] OSX

- [ ] [PLATFORM]:

### Functional tests

Adds new unit test.

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -20,6 +20,7 @@ import proseco.characteristics as ACA
 from astropy.table import Column, Table
 from chandra_aca.star_probs import guide_count
 from chandra_aca.transform import mag_to_count_rate, snr_mag_for_t_ccd, yagzag_to_pixels
+from cxotime import CxoTime
 from jinja2 import Template
 from proseco.catalog import ACATable
 from proseco.core import MetaAttribute
@@ -1173,6 +1174,7 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}{t_ccd_eff_acq_msg}""
         self.check_dither()
         self.check_fid_count()
         self.check_include_exclude()
+        self.check_t_ccd_acq()
 
     def check_guide_overlap(self):
         """
@@ -1350,6 +1352,25 @@ Predicted Acq CCD temperature (init) : {self.t_ccd_acq:.1f}{t_ccd_eff_acq_msg}""
                         msg = msg + f" halfwidths(s): {halfws}"
 
                     self.add_message("info", msg)
+
+    def check_t_ccd_acq(self):
+        """
+        Check that the fid lights are in the current search boxes by a check on t_ccd_acq.
+        These tolerances apply to observations after the 2022:294 safe mode.
+        """
+        if not self.is_ER and CxoTime(self.date).date > "2022:294":
+            if self.fids.t_ccd_acq <= -14.0:
+                self.add_message(
+                    "critical",
+                    f"Fid lights outside boxes (t_ccd_acq {self.fids.t_ccd_acq:.1f} <= -14.0)",
+                )
+                return
+            if self.fids.t_ccd_acq <= -13.5:
+                self.add_message(
+                    "caution",
+                    f"Fid lights near box edge (t_ccd_acq {self.fids.t_ccd_acq:.1f} <= -13.5)",
+                )
+                return
 
     def check_guide_count(self):
         """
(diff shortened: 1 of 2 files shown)


sparkles/core.py
```diff
@@ -1351,6 +1353,25 @@ def check_include_exclude(self):
 
                     self.add_message("info", msg)
 
+    def check_t_ccd_acq(self):
+        """
+        Check that the fid lights are in the current search boxes by a check on t_ccd_acq.
+        These tolerances apply to observations after the 2022:294 safe mode.
+        """
+        if not self.is_ER and CxoTime(self.date).date > "2022:294":
```
taldcroft, line 1361: Use the early-return pattern here to remove the need to have the entire rest of the method indented.
```suggestion
        if self.is_ER or CxoTime(self.date).date < "2022:294":
            return
```

sparkles/core.py
```diff
@@ -1351,6 +1353,25 @@ def check_include_exclude(self):
...
+            if self.fids.t_ccd_acq <= -14.0:
```
taldcroft, line 1362: Since you use `self.fids.t_ccd_acq` four times, define a short-cut like `t_ccd_acq` at this point and use it. Don't Repeat Yourself!

sparkles/core.py
```diff
@@ -1351,6 +1353,25 @@ def check_include_exclude(self):
...
+                self.add_message(
+                    "critical",
+                    f"Fid lights outside boxes (t_ccd_acq {self.fids.t_ccd_acq:.1f} <= -14.0)",
+                )
+                return
+            if self.fids.t_ccd_acq <= -13.5:
```
taldcroft, line 1368: This is not the best place for the early-return pattern because it does not reduce complexity. Instead:
```suggestion
            elif self.fids.t_ccd_acq <= -13.5:
```

sparkles/core.py
```diff
@@ -1351,6 +1353,25 @@ def check_include_exclude(self):
...
+                self.add_message(
+                    "caution",
+                    f"Fid lights near box edge (t_ccd_acq {self.fids.t_ccd_acq:.1f} <= -13.5)",
+                )
+                return
```
taldcroft, line 1373:
```suggestion
```

sparkles/tests/test_checks.py
```diff
@@ -80,6 +80,66 @@ def test_n_guide_check_not_enough_stars():
     ]
 
 
+def test_or_too_cold():
```
taldcroft, line 83: DRY.

Rewrite this using `pytest.mark.parametrize("case", cases)` where `cases` is a list of dict with the relevant inputs and expected `messages`. Get used to using `parametrize` for a number of test cases that are variations on a theme.

taldcroft (changes requested):

sparkles/tests/test_checks.py
```diff
@@ -80,6 +80,66 @@ def test_n_guide_check_not_enough_stars():
...
+        dark=DARK40,
+        raise_exc=True
+    )
+    acar = ACAReviewTable(aca)
+    acar.check_t_ccd_acq()
+    assert acar.messages == []
+
+    aca = get_aca_catalog(
+        **mod_std_info(
+            n_fid=3, n_guide=5, obsid=5000, t_ccd_acq=-13.6, date="2023:001"
```
taldcroft, line 110: The `t_ccd_acq` should be some number with  more precision than `-13.6`, e.g. `-13.62312`. That will test the rounding in the output.

sparkles/tests/test_checks.py
```diff
@@ -80,6 +80,66 @@ def test_n_guide_check_not_enough_stars():
...
+    )
+    acar = ACAReviewTable(aca)
+    acar.check_t_ccd_acq()
+    assert acar.messages == [
+        {
+            "text": "Fid lights outside boxes (t_ccd_acq -14.1 <= -14.0)",
+            "category": "critical",
+        }
+    ]
+
```
taldcroft, line 141: Also needs a test for an ER obsid.

sparkles/tests/test_checks.py
jeanconn: Well, I suppose in this new era DRY is "encourage copilot not to repeat itself when writing tests".

taldcroft: And that can be done by training it with good examples in your own code. :smile:

jeanconn:
Closing this as not needed just now.

PR #201, opened 2023-12-09
Update tests for ska_sun change to position_accurate

jeanconn:
## Description

Update test expected values for small roll changes caused by change to default method for position in ska_sun.

### Unit tests

- [x] Mac

### Functional tests

No functional testing.


sparkles/tests/test_review.py
```diff
@@ -629,7 +630,7 @@ def test_roll_options_for_not_allowed_pitch():
     roll = nominal_roll(ra, dec, date)
     att0 = Quat([ra, dec, roll])
 
-    att = apply_sun_pitch_yaw(att0, pitch=40)
+    att = apply_sun_pitch_yaw(att0, pitch=40, time=date)
```
jeanconn, line 633: The new ska_sun brakes on the previous line of code here... which I think was wrong anyway.

taldcroft:
I would have thought we just use the original fast sun positions via a `fast_sun_position_method` fixture that gets applied to each failing test. This is useful as a regression test to make sure that all the other coupled changes (chandra_maneuver, chandra_aca etc) don't introduce a change here.

taldcroft:
Though it is probably worth having at least one test that uses the default sun position and then the fast method. The final `test_get_params()` looks like a good candidate to basically duplicate into `test_get_params_sun_position_default` and `test_get_params_sun_position_fast`.

jeanconn:
I figured that unless you plan to update the application (sparkles) default to use position_fast (a possibility but one we hadn't discussed) that this was a just one-and-done change to the regress data. And having the changes as diffs in github was useful.  If you want to keep the position_fast test data as a different functional test I'm not sure I understand the value of that long term.

sparkles/tests/test_review.py
jeanconn: And this change is needed either way... though I think new ska_sun with position_accurate should probably handle the time=None sun_ra=None sun_dec=None case better than an astropy.time ValueError.

taldcroft: This is a real bug which I have a fix for, PR coming.

taldcroft (approved):
OK.

PR #203, opened 2023-12-31
Factor checks out of ACAReviewTable class

taldcroft:
## Description

This is to factor out the sparkles checks from being `ACAReviewTable` methods to independent check functions that return a list of messages. The motivation is to decouple these catalog checks from sparkles and make it easier to run them from starcheck or elsewhere.

Key updates:
- Move checks from being methods of the `ACAReviewTable` class into functions in a new `checks.py` module.
  - The functions take arbitrary arguments and return `list[Message]`.
  - In the existing sparkles checks the first argument is always an `ACACheckTable` or subclass, but that is no longer a requirement. 
  - Future checks should be written with more specific arguments where that would make sense.
- New `Message` dataclass and move `MessagesList`  into new `messages.py` module.
- Factor out a new `ACACheckTable` class that provides the minimal functionality to perform checks. This separates out all the sparkles **review** functionality (making reports, roll angle optimization, etc) from the basic catalog checking functionality:
  - `messages` attribute to capture messages
  - Additional columns like `row` and `col` etc to facilitate checking the catalog
  - Some helper attributes.
- Update check tests to use both `ACAReviewTable` and `ACACheckTable`.

The refactoring part is going to be difficult for code review but the tests are solid.

## Requires
- https://github.com/sot/proseco/pull/390 (for type annotations)
- `masters` channel in order to pass tests that depend on sun position.

### Unit tests

- [x] Mac
```
(masters) ➜  sparkles git:(refactor-checks) git rev-parse HEAD
ee2f79e6e50c4f97c25163dbf9f132c591594db8
(masters) ➜  sparkles git:(refactor-checks) pytest            
===================================================== test session starts ======================================================
platform darwin -- Python 3.10.8, pytest-7.2.1, pluggy-1.0.0
rootdir: /Users/aldcroft/git, configfile: pytest.ini
plugins: timeout-2.1.0, anyio-3.6.2
collected 103 items                                                                                                            

sparkles/tests/test_checks.py ............................................................................               [ 73%]
sparkles/tests/test_find_er_catalog.py .....                                                                             [ 78%]
sparkles/tests/test_review.py ..................                                                                         [ 96%]
sparkles/tests/test_yoshi.py ....                                                                                        [100%]

===================================================== 103 passed in 34.39s =====================================================
```

- [x] Linux

### Functional tests

No functional testing.

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -19,7 +19,7 @@ from jinja2 import Template
 from proseco.core import MetaAttribute
 
 from sparkles import checks
-from sparkles.aca_checks_table import ACAChecksTable
+from sparkles.aca_check_table import ACACheckTable
 from sparkles.messages import Message, MessagesList
 from sparkles.roll_optimize import RollOptimizeMixin
 
@@ -568,7 +568,7 @@ def get_summary_text(acas):
     return "\n".join(lines)
 
 
-class ACAReviewTable(ACAChecksTable, RollOptimizeMixin):
+class ACAReviewTable(ACACheckTable, RollOptimizeMixin):
     # Whether this instance is a roll option (controls how HTML report page is formatted)
     is_roll_option = MetaAttribute()
     roll_options = MetaAttribute()
@@ -1033,7 +1033,7 @@ check_pos_err_guide = checks.acar_check_wrapper(checks.check_pos_err_guide)
 check_too_bright_guide = checks.acar_check_wrapper(checks.check_too_bright_guide)
 
 
-def check_catalog(acar: ACAChecksTable) -> None:
+def check_catalog(acar: ACACheckTable) -> None:
     """Perform all star catalog checks."""
     msgs: list[Message] = []
     for entry in acar:
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -9,21 +9,19 @@ import pickle
 import pprint
 import re
 import traceback
-from itertools import chain, combinations
 from pathlib import Path
 
 import chandra_aca
 import numpy as np
 import proseco
-import proseco.characteristics as ACA
-from astropy.table import Column, Table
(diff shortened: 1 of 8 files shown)


sparkles/core.py
```diff
@@ -1629,3 +1013,67 @@ def from_ocat(cls, obsid, t_ccd=-5, man_angle=5, date=None, roll=None, **kwargs)
...
+check_guide_geometry = checks.acar_check_wrapper(checks.check_guide_geometry)
+check_guide_is_candidate = checks.acar_check_wrapper(checks.check_guide_is_candidate)
+check_guide_overlap = checks.acar_check_wrapper(checks.check_guide_overlap)
+check_imposters_guide = checks.acar_check_wrapper(checks.check_imposters_guide)
+check_include_exclude = checks.acar_check_wrapper(checks.check_include_exclude)
+check_pos_err_guide = checks.acar_check_wrapper(checks.check_pos_err_guide)
+check_too_bright_guide = checks.acar_check_wrapper(checks.check_too_bright_guide)
+
+
+def check_catalog(acar: ACACheckTable) -> None:
```
jeanconn, line 1036: It would make a little more sense to me to have check_catalog in checks, with an interface so that I could get the messages without writing back into the ACAReviewTable if desired.  But I'm not passionate about it.

taldcroft: The idea was that if some other application wants to use the checks from `checks.py`, then that app has its own version of `check_catalog` that calls some or all of the `checks.py` checks. So the `check_catalog()` function in `sparkles.core` is specifically *only* for sparkles, but the checks in the `checks` module are somewhat public and modular.

Of course another application would be free to copy the current `check_catalog` with essentially no modification except the last few lines.

jeanconn (approved):

PR #205, opened 2024-02-13
Refactor ruff into a base and package-specific

taldcroft:
## Description

This refactosr the ruff config into a base file which goes in `skare3/templates` and an extension to handle package-specific updates to the rules.

This uses the `ruff-base.toml` file in https://github.com/sot/skare3/pull/1269 and provides a demonstration of per-project extensions in `ruff.toml`.

## Interface impacts

None

### Unit tests
No code changes.

### Functional tests

`ruff .` and `ruff format` both succeed in the local repo.

### DIFF
--- a/ruff-base.toml
+++ b/ruff-base.toml
@@ -1,4 +1,4 @@
-# Copied originally from pandas
+# Copied originally from pandas. This config requires ruff >= 0.2.
 target-version = "py310"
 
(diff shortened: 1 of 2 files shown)


taldcroft:
I merged this without review since this doesn't impact code and mostly it's just me that cares about ruff.

PR #216, opened 2024-11-04
Update star mag, mag_err from AGASC

taldcroft:
## Description

This allows updating the ACA catalog `mag` and `mag_err` columns when reading catalogs from a proseco pickle file.

The driving use case is prelim review of obsid 30580 in NOV1124. This special observation had a pathfinder which resulted in updates to the AGASC that were not yet reflected in the prelim catalogs.

## Interface impacts

None, this PR makes no changes by default.

### Unit tests

- [ ] Mac

- [ ] [PLATFORM]:

### Functional tests

Obsid 30580
#### With this PR
image

#### Master
image

### DIFF
--- a/sparkles/core.py
+++ b/sparkles/core.py
@@ -6,17 +6,20 @@ Preliminary review of ACA catalogs selected by proseco.
 
 import gzip
 import io
+import itertools
 import pickle
 import pprint
 import re
 import traceback
 from pathlib import Path
 
+import agasc
 import chandra_aca
 import numpy as np
 import proseco
 from astropy.table import Table
 from jinja2 import Template
+from proseco.catalog import ACATable
 from proseco.core import MetaAttribute
 
 from sparkles import checks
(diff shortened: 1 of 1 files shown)


jeanconn:
I like this as a convenience, but maybe could use more convincing that it is necessary.  For similar cases we've used proseco to reselect with all stars force-include.

taldcroft:
Yup, it is a convenience. For convincing:
- Reduces brainload and time since you don't have to open a notebook or IPython and write code.
- Works easily on an entire weekly schedule.
- I already wrote and tested the code and it does not substantially complicate the code and API.

jeanconn:
Is this PR on hold for now or?

taldcroft:
> Is this PR on hold for now or?

Nope. I am still hoping that you are convinced this is useful based on the points I mentioned.

jeanconn:
Gotcha.  I thought for "Reduces brainload and time since you don't have to open a notebook or IPython and write code." that this PR made things worse as the one time we tried to use it, the subtleties of mag_err handling in proseco meant that we weren't on the same page with the FOT and the confusion was worse than brainload.

jeanconn:
And I thought that you'd agreed that just using proseco to reselect was probably safer and simpler.  Alternatively we could probably remove that special mag_err handling in proseco.

taldcroft:
Ok.

PR #220, opened 2025-01-04
Fix roll optimize deepcopy issue and fix warning in tests

taldcroft:
## Description

This fixes a [problem seen in testing of 2025.1rc2](https://chandramission.slack.com/archives/G01LN40AXPG/p1735926353950679), with the root cause being a changed introduced by  https://github.com/astropy/astropy/pull/15871.

It also makes a non-impacting change in tests to use `Table.pformat()` instead of `Table.pformat_all()`, since they are equivalent in astropy >= 7.0.

## Interface impacts

None

### Unit tests

- [x] Mac
```
ska3-flight-2025.0rc2) ➜  sparkles git:(fix-roll-optimize-deepcopy-issue) git rev-parse --short HEAD
c003f24
(ska3-flight-2025.0rc2) ➜  sparkles git:(fix-roll-optimize-deepcopy-issue) pytest                    
================================================= test session starts =================================================
platform darwin -- Python 3.12.8, pytest-8.3.4, pluggy-1.5.0
rootdir: /Users/aldcroft/git
configfile: pytest.ini
plugins: anyio-4.7.0, timeout-2.3.1
collected 104 items                                                                                                   

sparkles/tests/test_checks.py ............................................................................      [ 73%]
sparkles/tests/test_find_er_catalog.py .....                                                                    [ 77%]
sparkles/tests/test_review.py ...................                                                               [ 96%]
sparkles/tests/test_yoshi.py ....                                                                               [100%]

================================================ 104 passed in 29.43s =================================================
```

Independent check of unit tests by Javier
- [x] Linux
```
(ska3-flight-2025.0rc4) jgonzale sparkles $ pytest sparkles/
============================================================= test session starts =============================================================
platform linux -- Python 3.12.8, pytest-8.3.4, pluggy-1.5.0
rootdir: /proj/sot/ska/jgonzalez/git
configfile: pytest.ini
plugins: anyio-4.7.0, timeout-2.3.1
collected 104 items                                                                                                                           

sparkles/tests/test_checks.py ............................................................................                              [ 73%]
sparkles/tests/test_find_er_catalog.py .....                                                                                            [ 77%]
sparkles/tests/test_review.py ...................                                                                                       [ 96%]
sparkles/tests/test_yoshi.py ....                                                                                                       [100%]

======================================================= 104 passed in 80.71s (0:01:20) ========================================================
(ska3-flight-2025.0rc4) jgonzale sparkles $ git rev-parse --short HEAD
c003f24
```

### Functional tests

No functional testing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.2.2
+  rev: v0.8.5
   hooks:
     # Run the linter.
     - id: ruff
--- a/sparkles/roll_optimize.py
+++ b/sparkles/roll_optimize.py
@@ -5,7 +5,6 @@ Roll optimization during preliminary review of ACA catalogs.
 """
 
 import warnings
-from copy import deepcopy
 
 import numpy as np
 import ska_sun
@@ -367,8 +366,19 @@ class RollOptimizeMixin:
 
         att_targ = self.att_targ
 
-        # Special case, first roll option is self but with obsid set to roll
-        acar = deepcopy(self)
+        # Special case, first roll option is self but with obsid set to roll. Note that
+        # the "copy" from the line below is not a full deep copy of the table meta
+        # because it runs the __init__ method. This initializes the MetaAttributes to
+        # their default states instead of copying them. This is *required* for this code
+        # to work.
+        #
+        # As a development note, this used to be deepcopy(self), but in astropy 7.0 the
+        # implementation changed so that the meta dict was explicitly deepcopied after
+        # the __init__ call. This caused the MetaAttributes to be set to the original
+        # self values and the code to fail. The line below is the astropy 6.0 code for
+        # implementing deepcopy().
+        acar = self.copy(copy_data=True)
+
         check_catalog(acar)
         acar.is_roll_option = True
         roll_options = [
--- a/sparkles/roll_optimize.py
+++ b/sparkles/roll_optimize.py
@@ -377,6 +377,11 @@ class RollOptimizeMixin:
         # the __init__ call. This caused the MetaAttributes to be set to the original
         # self values and the code to fail. The line below is the astropy 6.0 code for
         # implementing deepcopy().
+        #
+        # See also https://github.com/sot/sparkles/pull/220/files#r1903130172 for more
+        # discussion and some clarification (maybe). One idea is to make do
+        # deepcopy(self) and not re-run check_catalog(); OR, make a brand new catalog
+        # from self.call_args.
         acar = self.copy(copy_data=True)
 
         check_catalog(acar)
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.2.2
+  rev: v0.8.5
```
javierggt, line 4: 0.8.6 was released two hours ago :)

sparkles/roll_optimize.py
```diff
@@ -367,8 +366,19 @@ def get_roll_options(
...
+        # because it runs the __init__ method. This initializes the MetaAttributes to
+        # their default states instead of copying them. This is *required* for this code
+        # to work.
+        #
+        # As a development note, this used to be deepcopy(self), but in astropy 7.0 the
+        # implementation changed so that the meta dict was explicitly deepcopied after
+        # the __init__ call. This caused the MetaAttributes to be set to the original
+        # self values and the code to fail. The line below is the astropy 6.0 code for
+        # implementing deepcopy().
+        acar = self.copy(copy_data=True)
```
javierggt, line 380: is this behavior going to be fixed upstream? I would expect that meta is copied if I were to call `copy.deepcopy`.

Of course, this does not matter for this PR. Just asking.

javierggt (approved):
The changes make sense to me and I verified that tests pass.

sparkles/roll_optimize.py
taldcroft: Meta is indeed deepcopied for a `deepcopy()` call, so astropy is doing the right thing. The copying of `meta` is actually the *problem* here, because the sparkles roll optimize code expects the `MetaData` attributes to be initialized to the default values, not to inherit from the original.

If you are still confused, this is why it took me several hours to really get to the bottom of this. The root cause is that the `ACAReviewTable` class is a bit unorthodox in that it sets the `MetaData` properties explicitly in the `__init__`. This introduces an ambiguity about copying: should the copy get the initial value from `__init__` or the copied value? Sparkles is expecting the initial value.

Overall, I agree this is a bit of a mess.

.pre-commit-config.yaml
taldcroft: Hah.

sparkles/roll_optimize.py
javierggt: I wonder if there is a way to make it so the values are set in `Table.__new__`, something like dding an argument that is only used by `Table.__new__`, that is `False` by default, but `Table.__deepcopy__` sets it to `True` when constructing the copy. I think that would run before `uper().__init_()`

taldcroft: @javierggt - I'm not quite following you. What would this new argument be controlling? It also sounds like you are implying an upstream change to `Table.__deepcopy__` for this, but that doesn't quite make sense. Of course the subclass in sparkles can override `__deepcopy__` if needed, but actually `deepcopy()` is currently behaving like I would expect (making an exact full copy).

taldcroft: Maybe a better thing would be to punt on making the copy and instead just reselect the catalog using the original `call_args` and hope that works.

taldcroft: Final discussion, consider changing above to a deepcopy and then do not re-run check_catalog().

